PR #23, opened 2025-06-09
Add NCS manifest

trond-snekvik:
Incoming additions around cryptography are significantly easier to implement on Nordic devices when using NCS than with raw zephyr. This adds an ncs manifest, so that we can run both.

### DIFF
--- a/west-ncs.yml
+++ b/west-ncs.yml
@@ -1,0 +1,20 @@
+manifest:
+  projects:
+    - name: nrf
+      revision: v3.0.1
+      url: http://github.com/nrfconnect/sdk-nrf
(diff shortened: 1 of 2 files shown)


west-ncs.yml
```diff
@@ -0,0 +1,6 @@
+manifest:
+  projects:
+    - name: nrf
+      revision: v3.0.1
+      url: http://github.com/nrfconnect/sdk-nrf
+      import: true
```
sam-golioth, line 6: Do we want to add allowlist to limit the number of repositories we clone?

mniestroj (approved):

PR #24, opened 2025-06-10
Certificate management

trond-snekvik:
Adds the certificate module, its required types, and its required kconfig values, but holds off on propagating this to the user facing API until the missing glue code is available.


src/cert.c
```diff
@@ -0,0 +1,266 @@
...
+    {
+        return err;
+    }
+
+    device.certificate = *cert;
+
+    return 0;
+}
+
+int cert_server_set(const struct pouch_cert *certbuf)
```
hasheddan, line 145: In practice this is almost certainly going to be a certificate chain, rather than a single leaf, as most `pouch` server implementations will leverage intermediate CA certs. We'll need to either deliver the certificate chain as concatenated PEM, or use a container format (e.g. PKCS#7) to deliver the bundled format, then handle parsing it here.

sam-golioth: I wonder if this is an opportunity to "sponsor" an intern to implement multiple certs upstream in MbedTLS. It wouldn't be ready for us soon enough, but it feels like getting support upstream is the right long term solution.

src/cert.c
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    memcpy(serial, server_cert.serial.data, server_cert.serial.len);
+
+    return server_cert.serial.len;
+}
+
+int pouch_device_certificate_ref_get(uint8_t *cert_ref, size_t len)
+{
+    if (!cert_is_valid(&device.certificate) || len > sizeof(device.ref))
```
sam-golioth, line 248:
```suggestion
    if (!cert_is_valid(&device.certificate) || len < sizeof(device.ref))
```

src/cert.c
```diff
@@ -0,0 +1,266 @@
...
+    {
+        return NULL;
+    }
+
+    return device.ref;
+}
+
+void cert_server_key_get(struct pubkey *out)
+{
+    *out = server_cert.pubkey;
```
sam-golioth, line 217: Any reason why we copy here but return a reference elsewhere (e.g. `pouch_device_certificate_get()`?

src/cert.c
```diff
@@ -0,0 +1,266 @@
...
+    {
+        return -EIO;
+    }
+
+    size_t hash_length;
+    status = psa_hash_finish(&hash, cert_ref, CERT_REF_LEN, &hash_length);
+    if (status != PSA_SUCCESS || hash_length != CERT_REF_LEN)
+    {
+        return -EIO;
+    }
```
sam-golioth, line 102: We're only calling `psa_hash_update()` once, so we should be able to use the one-shot API `psa_hash_compute()`

src/cert.c
```diff
@@ -0,0 +1,266 @@
...
+} device;
+static struct
+{
+    struct pubkey pubkey;
+    struct
+    {
+        uint8_t data[CERT_SERIAL_MAXLEN];
+        size_t len;
+    } serial;
+} server_cert;
```
sam-golioth, line 28: Are we able to store public keys in PSA storage? If so, why keep them here also? They're public, so it's not a security consideration, but it feels like it could be duplicative if the keys are stored in PSA for other reasons already (e.g. if a device is using an accelerator that has key storage).

src/cert.c
```diff
@@ -0,0 +1,266 @@
...
+
+static inline bool cert_is_valid(const struct pouch_cert *cert)
+{
+    return cert != NULL && cert->der != NULL && cert->size > 0;
+}
+
+static int parse_x509_cert(const struct pouch_cert *cert, mbedtls_x509_crt *out)
```
sam-golioth, line 35: It's unfortunate we can't use PSA here. No action to take, just voicing my frustration 😅

src/cert.c
sam-golioth: Since we know the length of our chain, we could also just have an explicit intermediate cert in the API for now. In terms of flexibility it's the same as just having a single cert, but at least it matches what we'll need for our implementation of the pouch server.

src/cert.c
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    device.certificate = *cert;
+
+    return 0;
+}
+
+int cert_server_set(const struct pouch_cert *certbuf)
+{
+    psa_key_attributes_t attrs = PSA_KEY_ATTRIBUTES_INIT;
```
sam-golioth, line 147: We don't appear to ever use this.

src/Kconfig
```diff
@@ -19,17 +19,79 @@ config POUCH_BLOCK_SIZE
...
     Select the encryption method to use for uplink Pouch transfers.
 
+config POUCH_ENCRYPTION_SAEAD
+  bool "Streaming AEAD"
+  select PSA_WANT_ALG_ECDSA
+  select PSA_WANT_ALG_ECDH
+  select PSA_WANT_ALG_HKDF
+  select PSA_WANT_ALG_STREAM_CIPHER
+  select PSA_WANT_ALG_SHA_256
+  select PSA_WANT_ALG_SHA_384
```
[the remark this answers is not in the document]
trond-snekvik: The server cert uses SHA384 for the digest :)

src/cert.c
trond-snekvik: Actually, we do want to be able to copy less than the full ref size for the 6 byte short ref

PR #26, opened 2025-06-13
SAEAD Sessions

trond-snekvik:
Introduces a session module for managing pouch sessions and their encrypted pouches. The session scope holds the following primitives:

- Session initiator
- Session ID
- Session ID format
- Session key
- Encryption algorithm

This module implements session ID and key generation, encryption, decryption as well as the session life cycle.

The session scope will be used to implement both uplink and downlink sessions, but since this is the core of the SAEAD encryption implementation (and since the lack of integration makes it not conflict with other ongoing work as much), I have split it out into its own PR.

### DIFF
--- a/src/saead/session.c
+++ b/src/saead/session.c
@@ -8,13 +8,19 @@
 #include <sys/types.h>
 #include <psa/crypto.h>
 #include <zephyr/sys/byteorder.h>
+#include <zephyr/sys/util.h>
 #include <zephyr/sys/base64.h>
 
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(saead_session, LOG_LEVEL_DBG);
 
+/**
+ * String length required to base64 encode a buffer of a given length, excluding the 0 terminator.
+ */
+#define BASE64_STRLEN(buflen) (4 * DIV_ROUND_UP(buflen, 3))
+
 #define NONCE_LEN 12
-#define INFO_MAX_LEN 34
+#define INFO_MAX_LEN (12 + BASE64_STRLEN(SESSION_ID_LEN) + 1)
 #define SAEAD_KEY_SIZE 32
 
 #define SESSION_KEY_TYPE(alg) \
@@ -63,9 +69,10 @@ int session_id_generate(struct session_id *id)
 
 static ssize_t session_key_info_build(const struct session_id *id,
                                       psa_algorithm_t algorithm,
+                                      uint8_t max_block_size_log,
                                       char *buf)
 {
-    char session_id[SESSION_ID_LEN * 2];
+    char session_id[BASE64_STRLEN(SESSION_ID_LEN) + 1];
     size_t id_len = 0;
 
     int err = base64_encode(session_id,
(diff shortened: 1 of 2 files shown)


src/saead/session.h
```diff
@@ -0,0 +1,91 @@
...
+};
+
+static inline bool session_id_is_equal(const session_id_t *a, const session_id_t *b)
+{
+    return a->type == b->type && a->initiator == b->initiator
+        && memcmp(&a->value, &b->value, sizeof(a->value)) == 0;
+}
+
+/** Initialize session */
+void session_init(struct session *session, psa_algorithm_t algorithm);
```
sam-golioth, line 70: Is the intention that `session` should be pre-populated with some data, or the application just passing in a piece of memory to be initialized?

src/saead/session.c
```diff
@@ -0,0 +1,321 @@
...
+
+    uint8_t nonce[NONCE_LEN];
+    nonce_generate(session, POUCH_ROLE_DEVICE, nonce);
+
+    LOG_INF("Session key: %d", session->key);
+
+    struct pouch_bufview plaintext;
+    pouch_bufview_init(&plaintext, block);
+
+    size_t plaintext_len = pouch_bufview_read_be16(&plaintext);
```
sam-golioth, line 214: Should we make a `_get_len()` function for blocks?

src/saead/session.c
```diff
@@ -0,0 +1,321 @@
...
+    if (plaintext_len <= AUTH_TAG_LEN || plaintext_len != pouch_bufview_available(&plaintext))
```
sam-golioth, line 215: Why does the plaintext need to be at least the length of an auth tag?

src/saead/session.c
sam-golioth, line 1: I think it would be "nice" to have an implementation of SAEAD that was separate from pouch, that just took in a key and initial AD (NULL in this case). Then this file would use it. But it's really not that much cleaner, it just gives us a fun thing to talk about and show off as a separate piece of technology, so consider this comment optional (more optional than usual 😅)

src/saead/session.c
```diff
@@ -0,0 +1,321 @@
...
+    status = psa_key_derivation_output_key(&key_attributes, &operation, &key);
+    if (status != PSA_SUCCESS)
+    {
+        LOG_ERR("Failed key derivation: %d", status);
+        goto exit;
+    }
+
+exit:
+    // Release the derivation context. Does not invalidate the key.
+    (void) psa_key_derivation_abort(&operation);
```
sam-golioth, line 164: Does this clear the memory holding the intermediate secrets?

sam-golioth (commented):
Very clean! Great stuff @trond-snekvik

A few comments, but overall love it.

src/saead/session.h
```diff
@@ -0,0 +1,91 @@
...
+        {
+            /** Randomized session tag */
+            uint8_t tag[SESSION_ID_TAG_LEN];
+            /** Sequence number for this session */
+            uint64_t seqnum;
+        } sequential;
+        /** Session random ID */
+        uint8_t random[SESSION_ID_LEN];
+    } value;
+} session_id_t;
```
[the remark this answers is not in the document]
trond-snekvik: Just a replay check on the sequence number. It's mostly exposed just to let the caller allocate it on the stack

src/saead/session.h
```diff
@@ -0,0 +1,91 @@
...
+        pouch_id_t id;
+        uint32_t block_index;
+        uint8_t ad[AD_LEN];
+    } pouch;
+};
+
+static inline bool session_id_is_equal(const session_id_t *a, const session_id_t *b)
+{
+    return a->type == b->type && a->initiator == b->initiator
+        && memcmp(&a->value, &b->value, sizeof(a->value)) == 0;
```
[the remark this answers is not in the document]
trond-snekvik: There could arguably be garbage in the padding though

src/saead/session.c
```diff
@@ -0,0 +1,324 @@
...
+    return key;
```
hasheddan, line 168: We return the key to the caller here, but do we actually ever set it on the `session`? Is the caller responsible for setting the session key? I'm not necessarily opposed to exposing the PSA key ID outside of this module, but it does feel somewhat leaky given that we manage its lifecycle internally (e.g. ending session destroys key).

trond-snekvik: We could pass in the whole session here to set the session key field, but I chose not to, as it makes it a bit hard to reason about which fields are expected to be populated and which will be set in the session object in this function. The session owners for uplink and downlink will be setting this, as well as the rest of the arguments in the session struct as soon as this passes

sam-golioth (approved):
Left a couple more comments, but non-blocking.

src/saead/session.c
hasheddan: Gotcha -- do you think it makes sense to remove the `algorithm` parameter from the `session_init` if that is the case?

hasheddan (approved):

src/saead/session.h
trond-snekvik: Removed this function altogether :)

PR #32, opened 2025-07-03
ci: add coverity workflow

sam-golioth:


.github/workflows/coverity.yml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  schedule:
+    # Run workflow every Sunday
+    - cron: "00 0 * * 0"
+  workflow_dispatch:
+
+jobs:
+  coverity:
+    runs-on: ubuntu-24.04
+    container: golioth/golioth-coverity-base:f08df75
```
mniestroj, line 12:
```suggestion
    container: golioth/golioth-coverity-base:89df175
```

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,54 @@
...
+      - uses: actions/checkout@v4
+        with:
+          path: pouch
+      - name: Init and update west
+        run: |
+          west init -l pouch --mf west.yml
+          west update --narrow -o=--depth=1
+      - name: Install pip packages
+        run: |
+          pip3 install \
```
mniestroj, line 24:
```suggestion
          uv pip install \
```

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,54 @@
...
+            -r deps/zephyr/scripts/requirements-base.txt        \
+            -r deps/zephyr/scripts/requirements-build-test.txt  \
+            -r deps/zephyr/scripts/requirements-run-test.txt    \
+            -r pouch/requirements.txt
+
+          pip3 install            \
```
mniestroj, line 30:
```suggestion
          uv pip install            \
```

.github/workflows/coverity.yml
```diff
@@ -0,0 +1,54 @@
+name: "Coverity"
+
+on:
+  schedule:
+    # Run workflow every Sunday
+    - cron: "00 0 * * 0"
```
mniestroj, line 6: Why do we want to run it every Sunday, instead of after each PR?

mniestroj (commented):
Should we just put all the Python dependencies into Docker image instead?

If not, then it is probably a good idea to migrate to `uv pip` already, as that will be pulled into newer versions of Docker image anyway (there is `golioth/golioth-zephyr-base:0.17.0-SDK-v0` as base image, which already contains `uv pip`).

sam-golioth:
> Should we just put all the Python dependencies into Docker image instead?
> 
> If not, then it is probably a good idea to migrate to `uv pip` already, as that will be pulled into newer versions of Docker image anyway (there is `golioth/golioth-zephyr-base:0.17.0-SDK-v0` as base image, which already contains `uv pip`).

I didn't realize that we had updated this image with `uv` as well, but I see know that it picked it up automatically! I'll change these to use `uv pip`.

hasheddan (approved):

PR #33, opened 2025-07-09
Encrypted uplink

trond-snekvik:
Adds uplink SAEAD sessions, and a crypto_saead.c module.
This adds SAEAD support to the pouch header and config structure, adds some lifecycle functions to the crypto modules and shuffles the configuration ownership around a bit so that it makes more sense.


trond-snekvik:
I believe I've found a solution to the configuration issue. It prevents users from adding their own MBED_TLS_USER_CONFIG file, but aside from this, I think it's a good enough solution until we contribute something permanent to upstream.

PR #34, opened 2025-07-10
Downlink decryption

trond-snekvik:
Adds downlink decryption and session management. The saead downlink session starts when we successfully parse the pouch header, which initializes the required decryption context, allowing decryption to run on the incoming pouches block by block.


src/saead/downlink.c
```diff
@@ -0,0 +1,161 @@
...
+
+        if (!uplink_session_matches(id, max_block_size_log, algorithm))
+        {
+            // The server claims to use our uplink's session, but it doesn't
+            LOG_ERR("Session reuse failed: No match");
+            return -EBADMSG;
+        }
+
+        // We can make a copy of the uplink session's key instead of deriving it again:
+        session_key = uplink_session_key_copy(DOWNLINK_KEY_USAGE);
```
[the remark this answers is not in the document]
trond-snekvik: That's correct, yes. This occupies a separate slot in the key storage module, but in my opinion, the complexity associated with creating a reference counting-esque mechanism outweighs this drawback.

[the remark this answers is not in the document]
hasheddan: @trond-snekvik that's the same path I was considering, and I agree that, especially for the time-being, this is a lower complexity and safer option. If we find ourselves dealing with with slot exhaustion at some point in the future I think we can likely make this configurable 👍🏻 

(side note: in theory this could be pushed to the PSA implementation in that it could manage copies by ref counting rather than actually occupying a multiple physical slots)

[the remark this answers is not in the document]
trond-snekvik: > (side note: in theory this could be pushed to the PSA implementation in that it could manage copies by ref counting rather than actually occupying a multiple physical slots)

I'm hoping that this is  the case, but the spec doesn't specify as far as I can tell

src/downlink.c
```diff
@@ -119,6 +119,13 @@ static size_t pouch_downlink_parse_header(struct pouch_bufview *v)
     LOG_DBG("Header version %d", (int) header.version);
     LOG_DBG("Payload len %d", (int) header_len);
```
sam-golioth, line 120: Can we add a `LOG_DBG` for the encryption type?

src/saead/downlink.c
sam-golioth, line 1: Thoughts on encrypt/decrypt for this module instead of uplink/downlink?

EDIT: After reading through the rest of the PR, I'm less sure of this, as some of the session establishment/replay protection does feel specific to uplink vs downlink. Still, I'd love to see a "generic" SAEAD encrypt/decrypt implementation, with pouch-specific sessions layered on top. But I don't love it enough to hold up this PR.

src/saead/downlink.c
```diff
@@ -0,0 +1,161 @@
+#include "downlink.h"
```
sam-golioth, line 1: Missing copyright/license header

src/saead/downlink.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
```
sam-golioth, line 1: Missing copyright/license header

src/saead/downlink.c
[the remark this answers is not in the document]
sam-golioth: Doing it in the PSA implementation also provides a clean(er) way for this to be added later or implemented downstream if its something that customers hit in their particular application (e.g. if they are using more keyslots for other parts of their application).

src/saead/downlink.c
```diff
@@ -0,0 +1,161 @@
...
+    /**
+     * Highest sequence number we've seen the server create a session with.
+     * If replay protection isn't enabled, we can still use this to ensure that the server hasn't
+     * sent the same message since our last power cycle. This isn't proper replay protection, as
+     * it'll get wiped when this device power cycles, but it's a cheap guard against low effort
+     * replay attacks.
+     */
+    uint64_t seqnum;
+    /**
+     * Highest pouch ID we've seen the server send with. Is only updated once at least one block of
```
sam-golioth, line 27:
```suggestion
     * Highest pouch ID we've seen the server send with in the current session. Is only updated once at least one block of
```

src/saead/downlink.c
```diff
@@ -0,0 +1,161 @@
...
+        if (id->type != SESSION_ID_TYPE_SEQUENTIAL)
+        {
+            // The server can only use our session if it's a sequential session ID
+            LOG_ERR("Session reuse failed: ID not sequential");
+            return -EBADMSG;
+        }
+
+        if (!uplink_session_matches(id, max_block_size_log, algorithm))
+        {
+            // The server claims to use our uplink's session, but it doesn't
```
sam-golioth, line 87:
```suggestion
            // The server claims to use our uplink's session, but it doesn't match
```

src/crypto_saead.c
```diff
@@ -38,6 +39,59 @@ int crypto_init(const struct pouch_config *config)
...
+                                         ? PSA_ALG_GCM
+                                         : PSA_ALG_CHACHA20_POLY1305,
+                                     session->max_block_size_log,
+                                     pkey);
+    if (err)
+    {
+        return err;
+    }
+
+    return downlink_pouch_start(encryption_info->saead_info_m.pouch_id);
```
sam-golioth, line 92: In src/downlink.c, we have a function called `pouch_downlink_start()`. Can we prefix this (and the other downlink/uplink public APIs) with `saead_`? I think it will be easier to read.

src/crypto_saead.c
```diff
@@ -38,6 +39,59 @@ int crypto_init(const struct pouch_config *config)
...
+        memcpy(id.value.sequential.tag,
+               session->session_id_sequential_m.tag.value,
+               sizeof(id.value.sequential.tag));
+        id.value.sequential.seqnum = session->session_id_sequential_m.seq;
+    }
+
+    int err = downlink_session_start(&id,
+                                     session->algorithm_choice == session_info_algorithm_aes_gcm_m_c
+                                         ? PSA_ALG_GCM
+                                         : PSA_ALG_CHACHA20_POLY1305,
```
sam-golioth, line 84: Will this actually just work for ChaCha20-Poly1305 or is there more we would need to implement?

src/saead/downlink.c
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    return session_pouch_start(&downlink, id);
+}
+
+struct pouch_buf *downlink_block_decrypt(struct pouch_buf *block)
+{
+    struct pouch_buf *decrypted = session_decrypt_block(&downlink, block);
+    buf_free(block);
+    if (!decrypted)
```
sam-golioth, line 144:
```suggestion
    if (decrypted == NULL)
```

sam-golioth (commented):
Thanks @trond-snekvik! Mostly cosmetic feedback.

mniestroj (approved):

src/saead/downlink.c
trond-snekvik: These files having the same name as the corresponding files in the folder above this one also bugs me, but it does feel like it's the most accurate name, unfortunately. I think the entire module split here is a little awkward though, so there might be value in doing some refactoring here a bit later.

I think a generic pouch encrypt/decrypt module would mostly contain code from saead/session.c, but it could be a good start to create better modules here overall.

PR #36, opened 2025-07-15
transport: ble_gatt: add server cert chrc

mniestroj:
Add support for writing server cert as well as reading serial of already stored
server cert.


src/transport/ble_gatt/server_cert_characteristic.c
```diff
@@ -0,0 +1,156 @@
...
+                                 uint16_t len,
+                                 uint16_t offset,
+                                 uint8_t flags)
+{
+    struct golioth_ble_gatt_server_cert_ctx *ctx = attr->user_data;
+    bool is_first = false;
+    bool is_last = false;
+    const void *payload = NULL;
+    ssize_t payload_len =
+        golioth_ble_gatt_packetizer_decode(buf, len, &payload, &is_first, &is_last);
```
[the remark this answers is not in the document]
sam-golioth: From the [spec](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-61/out/en/host/attribute-protocol--att-.html#UUID-d24e0156-618e-728d-f19b-01349ced952b):
> The maximum length of an attribute value shall be 512 octets.

In practice, I've seen stacks ignore this limit (including, I believe, the Zephyr stack). But doing so is non-compliant so we should stick to 512 as our max. A certificate, and especially a certificate chain, can easily exceed that length, so we'll need to split it using the packetizer.

As for the [serial](https://github.com/golioth/pouch/pull/36/files/70de936089a2950d7667ec9be1fe4440749131be#r2207478373), I agree that it's not strictly required, but I'm leaning towards using it for consistency with the other Toothfairy characteristics.

sam-golioth (approved):
Thanks @mniestroj!

src/transport/ble_gatt/server_cert_characteristic.c
```diff
@@ -0,0 +1,156 @@
...
+
+    memcpy((void *) &ctx->cert.der[ctx->cert.size], payload, payload_len);
+    ctx->cert.size += payload_len;
+
+    if (is_last)
+    {
+        pouch_server_certificate_set(&ctx->cert);
+
+        free((void *) ctx->cert.der);
+        ctx->cert.der = NULL;
```
hasheddan, line 143: I don't think we can free here because because the call stack from [`pouch_server_certificate_set()`](https://github.com/golioth/pouch/blob/b40a69484fe69ba0897f425e6830b9c73fc11b5d/src/cert.c#L218) is:
- [`cert_server_set()`](https://github.com/golioth/pouch/blob/b40a69484fe69ba0897f425e6830b9c73fc11b5d/src/cert.c#L136)
- [`parse_x509_crt()`](https://github.com/golioth/pouch/blob/b40a69484fe69ba0897f425e6830b9c73fc11b5d/src/cert.c#L35)
- [`mbedtls_x509_crt_parse_der_nocopy()`](https://github.com/golioth/pouch/blob/b40a69484fe69ba0897f425e6830b9c73fc11b5d/src/cert.c#L44)

The use of the `nocopy` variant here means we are [transitioning ownership of the buffer to the CRT](https://github.com/Mbed-TLS/mbedtls/blob/8f548c9dba2578d3d045c3e5744e24ead4671cfc/include/mbedtls/x509_crt.h#L471C1-L475C62). We'll need to either pass a copy of the buffer here or avoid using the `nocopy` API.

sam-golioth: I think it's fine because the parsed cert is a stack variable. The relevant cert fields are copied into the static `server_cert` struct:

```
    uint8_t *pk = parsed_cert.pk.MBEDTLS_PRIVATE(pub_raw);
    size_t pk_len = parsed_cert.pk.MBEDTLS_PRIVATE(pub_raw_len);
    if (pk_len > sizeof(server_cert.pubkey.data))
    {
        LOG_ERR("Unexpected server public key size: %u", pk_len);
        goto exit;
    }

    if (parsed_cert.serial.len > sizeof(server_cert.serial.data))
    {
        LOG_ERR("Unexpected server certificate serial number size: %u", parsed_cert.serial.len);
        goto exit;
    }

    if (IS_ENABLED(CONFIG_POUCH_VALIDATE_SERVER_CERT))
    {
        err = authenticate_server_cert(&parsed_cert);
        if (err)
        {
            goto exit;
        }
    }

    memcpy(server_cert.pubkey.data, pk, pk_len);
    server_cert.pubkey.len = pk_len;

    memcpy(server_cert.serial.data, parsed_cert.serial.p, parsed_cert.serial.len);
    server_cert.serial.len = parsed_cert.serial.len;
```

trond-snekvik: @hasheddan I'm on my phone here, so this is a bit of a drive by comment, but doesn't mbedtls yield ownership of the buffer when the certificate is freed in this case? It has no way of knowing what heap the buffer should be freed from, so it can't actually reliably free the buffer. I was under the impression that this just meant that the buffer has to be valid for the lifecycle of the cert, as mbedtls won't make its own copy?

hasheddan: @sam-golioth @trond-snekvik yep you're right -- I missed that we were copying out the data and freeing before returning:
https://github.com/golioth/pouch/blob/b40a69484fe69ba0897f425e6830b9c73fc11b5d/src/cert.c#L190

Apologies for the noise here!

hasheddan (approved):

mniestroj:
Recent pushes were only including small cmake changes and rebase on main.

PR #38, opened 2025-07-18
Settings

sam-golioth:
This adds a `golioth_sdk` module and Settings as the first service. This iteration of the SDK is using a frontend/backend approach for the services. The backend handles the serialization and deserialization of the data, while the frontend provides different ways for the application to interact with the service. For example, this PR adds a callback based frontend for the settings service, but that could be swapped out for a frontend that loads the settings values directly into the Zephyr settings subsystem. It also provides an opportunity (hopefully not needed often) to switch out the backends as well, if we need different serialization/deserialization for different protocols. This architecture will be refined further as we add and expand additional services.

Resovles golioth/firmware-issue-tracker#810


golioth_sdk/settings.c
```diff
@@ -0,0 +1,182 @@
...
+    zcbor_tstr_put_lit(zse, "version");
+    if (settings_version >= 0)
+    {
+        zcbor_int64_put(zse, settings_version);
+    }
+    else
+    {
+        zcbor_nil_put(zse, NULL);
+    }
+    zcbor_map_end_encode(zse, 2);
```
sam-golioth, line 173: This PR doesn't handle reporting of settings errors. That will come in a follow on PR.

golioth_sdk/settings_callbacks.c
```diff
@@ -0,0 +1,191 @@
...
+            if (setting->type != value->type)
+            {
+                return -EFTYPE;
+            }
+
+            switch (setting->type)
+            {
+                case SETTING_VALUE_TYPE_INT:
+                    return setting->int_cb(value->int_val, setting->cb_arg);
+                    break;
```
trond-snekvik, line 71: I'm surprised the linter doesn't complain about these breaks

include/golioth/golioth.h
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright (c) 2025 Golioth
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+int golioth_sync_to_cloud(void);
```
trond-snekvik, line 7: Needs docstring.

This isn't blocking, but should it be? Perhaps the K_FOREVER timeout parameter should get elevated here? 

Also, we could potentially emit an event when this finishes, like `GOLIOTH_CLOUD_SYNC_COMPLETE`

golioth_sdk/dispatch.c
```diff
@@ -0,0 +1,89 @@
...
+{
+    STRUCT_SECTION_FOREACH(golioth_service, service)
+    {
+        if (NULL != service->uplink_cb)
+        {
+            service->uplink_cb();
+        }
+    }
+
+    pouch_uplink_close(K_FOREVER);
```
trond-snekvik, line 84: Seeing this in the code, it does feel a bit like this function is doing more than one thing, but I do think it's a better UX than forcing the user to call this separately, so IDK. Definitely feels like a gotcha that you have to finish your other uplink stuff before calling this to avoid sending multiple pouches.

lib/zcbor_utils/Kconfig
```diff
@@ -0,0 +1,2 @@
+config ZCBOR_UTILS
+    bool "Helper functions for interacting with CBOR maps"
```
trond-snekvik, line 2: Does this really need a kconfig variable?

golioth_sdk/settings.c
```diff
@@ -0,0 +1,182 @@
+/*
+ * Copyright (c) 2025 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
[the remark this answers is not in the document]
sam-golioth: No, the whole repo is [Apache 2.0 licensed](https://github.com/golioth/pouch/blob/main/LICENSE#L1). I missed the SPDX identifier in a couple files here, I'll update this PR to add them. It's also missing from several other files in the repo, I'll open a separate PR to add it.

golioth_sdk/settings_callbacks.c
sam-golioth: You mean because they're not aligned with the `case`? This is actually where the linter wants them (it will move them here if they're at another alignment). This formatting is pretty standard across most codebases I've seen, but I personally don't like it. The logic for this alignment (I think) is that `break` is always part of the code block it's breaking out of (e.g. in a while loop, it's always indented at the same level as the loop body).

My personal preference is something like:

```
switch (foo)
{
    case bar:
    {
        do_something;
    }
    break;
    
    case baz:
    {
        do_something_else;
    }
    break;
}
```

It gives your brain the same cues as all the other scopes (functions, structs, if, while, etc.), but this is another case of clang-format not being as configurable as we want it to be: https://github.com/llvm/llvm-project/issues/58669

golioth_sdk/dispatch.c
sam-golioth: Yeah, that's a good point. I hadn't thought about that because of the roundabout way I came around to this from the discussion at breakout. I think there's good arguments for either way, so let's start with this and see how it feels as we go through beta? I think seeing how stream fits into the SDK will also help inform this. It should be a simple change to move this close call up to the application.

lib/zcbor_utils/Kconfig
sam-golioth: The idea with this library is that users can include it and use it in their application without needing the Golioth SDK turned on. The other option would be adding into the golioth_sdk folder and making it private to the SDK. I'm leaning towards the current solution but very open to changing it.

golioth_sdk/settings_callbacks.c
sam-golioth: Yeah that compile fines, I'll remove the `break`s. I swear I'm good at coding 😅

include/golioth/golioth.h
trond-snekvik: I think so, yeah. I'm not really that into making pouch separate from golioth from the beginning here, but if we start mixing them, it doesn't seem likely that we'd ever be able to separate them.

include/golioth/settings_callbacks.h
```diff
@@ -0,0 +1,52 @@
...
+
+/// Same as @ref golioth_settings_register_int, but for type float.
+int golioth_settings_register_float(const char *setting_name,
+                                    golioth_float_setting_cb callback,
+                                    void *callback_arg);
+
+/// Same as @ref golioth_settings_register_int, but for type string.
+int golioth_settings_register_string(const char *setting_name,
+                                     golioth_string_setting_cb callback,
+                                     void *callback_arg);
```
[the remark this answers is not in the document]
sam-golioth: That is pretty dang cool. Yes, the runtime stuff is because it's pulled out of the SDK. Linker arrays are a good fit here, but I'm a little concerned about deviating too far from the SDK. Let me noodle a bit.

include/golioth/golioth.h
sam-golioth: I added a docstring (and updated the settings docstrings to use the block comment format). I'm going to defer adding a timeout parameter and event/callback until we make the change to run this on its own thread (in a different PR).

include/golioth/settings_callbacks.h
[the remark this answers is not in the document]
sam-golioth: Fair points. I also really like the linker array pattern a lot. I've used it before outside of Zephyr, it's actually pretty straightforward to implement, but it does require customers to make changes to their linker scripts, which makes it harder to work out of the box. In the past, we've indexed heavily on the getting started experience, even if it's not a great representation of what developing in the SDK would look like, but I'm not sure we've struck the right balance.

With this split architecture, we also have the option of supporting both runtime and compile time registration, through different frontends, if we decide later that we don't want to force users to support linker arrays outside of Zephyr. So, I'll change this to use Zephyr's iterable sections, and we can address how to add support for other platforms later.

golioth_sdk/dispatch.h
```diff
@@ -0,0 +1,34 @@
...
+typedef void (*golioth_service_uplink_cb)(void);
+
+#define STREAM_ID_INVALID (-1)
+
+struct golioth_service
+{
+    const char *path;
+    int stream_id;
+    golioth_service_downlink_cb downlink_cb;
+    golioth_service_uplink_cb uplink_cb;
```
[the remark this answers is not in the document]
sam-golioth: Done!

include/golioth/settings_callbacks.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+#pragma once
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include <golioth/settings_types.h>
+
+/**
+ * Callback function types for golioth_settings_register_*
```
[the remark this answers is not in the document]
sam-golioth: Karma lol. (fixed)

PR #40, opened 2025-07-21
Example: Support certificates

trond-snekvik:
Adds support for certificates in the example. Certificates can both be loaded by mcumgr or be embedded in the firmware.


examples/ble_gatt/src/credentials.c
```diff
@@ -0,0 +1,194 @@
...
+    return psa_generate_random(output, output_len);
+}
+
+/** Load the raw private key data into PSA */
+static psa_key_id_t import_raw_pk(const uint8_t *private_key, size_t size)
+{
+    mbedtls_pk_context pk;
+    mbedtls_pk_init(&pk);
+    int err = mbedtls_pk_parse_key(&pk, private_key, size, NULL, 0, psa_rng_for_mbedtls, NULL);
+    if (err)
```
sam-golioth, line 23:
```suggestion
    if (err != 0)
```

examples/ble_gatt/src/credentials.c
```diff
@@ -0,0 +1,194 @@
...
+
+    psa_key_attributes_t attrs = PSA_KEY_ATTRIBUTES_INIT;
+
+    psa_set_key_algorithm(&attrs, PSA_ALG_ECDH);
+    psa_set_key_type(&attrs, PSA_KEY_TYPE_ECC_KEY_PAIR(PSA_ECC_FAMILY_SECP_R1));
+    psa_set_key_usage_flags(&attrs, PSA_KEY_USAGE_DERIVE);
+
+    psa_key_id_t key_id;
+    err = mbedtls_pk_import_into_psa(&pk, &attrs, &key_id);
+    if (err)
```
sam-golioth, line 37:
```suggestion
    if (err != 0)
```

examples/ble_gatt/src/credentials.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
```
sam-golioth, line 1: Missing copyright/license header

examples/ble_gatt/src/main.c
```diff
@@ -116,18 +118,35 @@ int main(void)
         return 0;
     }
 
-    LOG_DBG("Bluetooth initialized\n");
+    LOG_DBG("Bluetooth initialized");
+
+    struct pouch_config config;
+    err = load_certificate(&config.certificate);
+    if (err)
```
sam-golioth, line 125:
```suggestion
    if (err != 0)
```

examples/ble_gatt/src/credentials.c
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright (c) 2025 Golioth, Inc.
```
sam-golioth, line 2: Let's add the SPDX identifier as well

examples/ble_gatt/src/main.c
sam-golioth, line 1: Let's add a README for this application with instructions for loading credentials using SMP. I know there's been pushback against `mcumgr` because of the go dependency - we can recommend `smpmgr` from pip instead.

sam-golioth (changes requested):
We recently removed support for hardcoded credentials in the Firmware SDK, and I don't think we should be adding them here. The logic behind removing them is:
- It models improper design for users, who more often than we'd expect end up using code in production that we intend to only be available for samples, and then ask us for support
- It creates additional code that we need to maintain and test, but isn't (shouldn't be) going into customer applications.
- The delta between hardcoding credentials and loading them at runtime isn't that large - it's more of a mental barrier.
  - It's certainly more complex to use certificates than PSKs, but I'd argue that most of the complexity is involved in generating the keys - uploading the certificate to the device is a process that can be clearly described in just a few easy to follow steps.
  
I believe that most of the pushback from devrel has been around installing `go` and compiling `mcumgr`, rather than actually running through the steps to load the cert. I mentioned this in another comment, but we can use `smpmgr` installed via `pip` instead.

examples/ble_gatt/src/credentials.c
trond-snekvik: Small, bikeshed-shaped hill to die on, but I actually don't want to do this change, and I'd rather type out a response than leave it unaddressed.

The overwhelming majority of error checks in this repo and the firmware SDK uses `if (err)`, like Zephyr and Linux does. A similar ratio appears to be practiced for pointer checks, where we're using `if (!ptr)` more often than `if (ptr == NULL)`, particularly after allocations.

I recognize that this is a rather enraging response to a trivial suggestion in a PR, but unless it is in the style guide, it's a personal preference, IMO. I prefer `if (err)` because it reads more like English and follows the same convention as Zephyr (and the Linux Kernel, which is what our style guide defers to). I'm aware of the [CERT recommendation](https://wiki.sei.cmu.edu/confluence/display/c/EXP20-C.+Perform+explicit+tests+to+determine+success%2C+true+and+false%2C+and+equality), but for the error and NULL checks, I think readability and sticking with a familiar convention is more important. The scenarios listed as potential regressions in the recommendation are arguably rather contrived and extremely unlikely to manifest when we have a sane error return value convention. I absolutely agree with explicit equality checks for non-standard values like enums though, and would definitely consider it problematic if we used an implicit comparison to check for the new `GOLIOTH_SETTING_VALUE_TYPE_UNKNOWN`, for example.

trond-snekvik:
I would like to keep the hardcoded credentials as an alternative here, but I think all my arguments for it have already been presented, so I don't think I'll be able to come up with some rebuttal that could change your mind :smile: 

I think my viewpoint here is that we only have "don't do this at home"-solutions to this problem at the moment, so we might as well give users the foot gun with the fewest amounts of steps required to get started.

examples/ble_gatt/src/main.c
trond-snekvik: I tested smpmgr, but [this issue](https://github.com/intercreate/smpmgr/issues/51) prevents the file uploads from working, unfortunately. The certificate file is too large to be transferred in time for the implicit timeout, even when I set the timeout flag. Happy to add it in once that is resolved, as the installation and UX is really slick.

examples/ble_gatt/src/credentials.c
sam-golioth: Fair enough - it's not a hill I want to die on 🙂

examples/ble_gatt/src/credentials.c
```diff
@@ -0,0 +1,165 @@
...
+
+    LOG_INF("Read certificate (%d bytes)", size);
+
+    return 0;
+}
+
+void free_certificate(struct pouch_cert *cert)
+{
+    free((void *) cert->der);
+}
```
mniestroj, line 165: Do we want to update `size` here and/or assign `der` to `NULL`? This usually helps with maintaining information whether data should be valid or not (either in runtime or during debugging).

examples/ble_gatt/src/main.c
mniestroj: For future, we could try to wrap `smpmgr` into `west` command, choose proper MTU (if that needs to be set explicitly) and probably also integrate with our Python REST API to simplify provisioning. Just dropping the idea here.

mniestroj (approved):

hasheddan (approved):

examples/ble_gatt/src/main.c
sam-golioth: @JPHutchins When I enable the mcumgr parameters (`CONFIG_MCUMGR_GRP_OS_MCUMGR_PARAMS`), the warning you mention goes away, but the file transfer still fails if I don't manually set the MTU. This is on smpmgr v0.13.1.

Without specifying the mtu:
```
(golioth) samfriedman ~/golioth/bluetooth-gateway/bluetooth-gateway [main] $ smpmgr --port /dev/cu.usbmodem0010502274171 file upload ../../certs/playground-2f69a2-0000000001.key.der /lfs1/credentials/key.der
⠋ Connecting to /dev/cu.usbmodem0010502274171... OK
[11:43:58] ERROR    Timeout (2.5s) waiting for request header=Header(op=<OP.WRITE: 2>, version=<Version.V2: 1>, __init__.py:170
                    flags=<Flag.UNUSED: 0>, length=172, group_id=<GroupId.FILE_MANAGEMENT: 8>, sequence=1,
                    command_id=<FileManagement.FILE_DOWNLOAD_UPLOAD: 0>) version=<Version.V2: 1> sequence=1
                    smp_data=b"\n\x00\x00\xac\x00\x08\x01\x00\xa4clen\x18ycoff\x00ddataXy0w\x02\x01\x01\x04
                    D\xa0!\xe1\x12\\QM\n\x8c\xffO\x80\xf2D\xb9\xbb]\xc1\xa6\xc9\xeb\xc2\x8c\xb0\xd3Axb\xe2\xca>
                    \xa0\n\x06\x08*\x86H\xce=\x03\x01\x07\xa1D\x03B\x00\x04\xfa\x9a/\x15,\xf5\x1e;\x88@\xee\xa2
                    \xf3\x91\x07'\x91\xa1b\xafJ{:\xee\xfc\x8aZ\x00\x1b\x98\xa0}2t\xbc\x92\xf3^J\xa1\xfe\x8e\xd8
                    i.j:\xc2~-+\x1cO@|\xdc\x08\x97\x86\x00;5\x97\xecdnamex\x19/lfs1/credentials/key.der" off=0
                    data=b"0w\x02\x01\x01\x04
                    D\xa0!\xe1\x12\\QM\n\x8c\xffO\x80\xf2D\xb9\xbb]\xc1\xa6\xc9\xeb\xc2\x8c\xb0\xd3Axb\xe2\xca>
                    \xa0\n\x06\x08*\x86H\xce=\x03\x01\x07\xa1D\x03B\x00\x04\xfa\x9a/\x15,\xf5\x1e;\x88@\xee\xa2
                    \xf3\x91\x07'\x91\xa1b\xafJ{:\xee\xfc\x8aZ\x00\x1b\x98\xa0}2t\xbc\x92\xf3^J\xa1\xfe\x8e\xd8
                    i.j:\xc2~-+\x1cO@|\xdc\x08\x97\x86\x00;5\x97\xec" name='/lfs1/credentials/key.der' len=121
                    - __init__:170
           ERROR    Connection to device lost: TimeoutError - Timeout (2.5s) waiting for request         file_management.py:133
                    header=Header(op=<OP.WRITE: 2>, version=<Version.V2: 1>, flags=<Flag.UNUSED: 0>,
                    length=172, group_id=<GroupId.FILE_MANAGEMENT: 8>, sequence=1,
                    command_id=<FileManagement.FILE_DOWNLOAD_UPLOAD: 0>) version=<Version.V2: 1>
                    sequence=1
                    smp_data=b"\n\x00\x00\xac\x00\x08\x01\x00\xa4clen\x18ycoff\x00ddataXy0w\x02\x01\x01\
                    x04
                    D\xa0!\xe1\x12\\QM\n\x8c\xffO\x80\xf2D\xb9\xbb]\xc1\xa6\xc9\xeb\xc2\x8c\xb0\xd3Axb\x
                    e2\xca>\xa0\n\x06\x08*\x86H\xce=\x03\x01\x07\xa1D\x03B\x00\x04\xfa\x9a/\x15,\xf5\x1e
                    ;\x88@\xee\xa2\xf3\x91\x07'\x91\xa1b\xafJ{:\xee\xfc\x8aZ\x00\x1b\x98\xa0}2t\xbc\x92\
                    xf3^J\xa1\xfe\x8e\xd8i.j:\xc2~-+\x1cO@|\xdc\x08\x97\x86\x00;5\x97\xecdnamex\x19/lfs1
                    /credentials/key.der" off=0 data=b"0w\x02\x01\x01\x04
                    D\xa0!\xe1\x12\\QM\n\x8c\xffO\x80\xf2D\xb9\xbb]\xc1\xa6\xc9\xeb\xc2\x8c\xb0\xd3Axb\x
                    e2\xca>\xa0\n\x06\x08*\x86H\xce=\x03\x01\x07\xa1D\x03B\x00\x04\xfa\x9a/\x15,\xf5\x1e
                    ;\x88@\xee\xa2\xf3\x91\x07'\x91\xa1b\xafJ{:\xee\xfc\x8aZ\x00\x1b\x98\xa0}2t\xbc\x92\
                    xf3^J\xa1\xfe\x8e\xd8i.j:\xc2~-+\x1cO@|\xdc\x08\x97\x86\x00;5\x97\xec"
                    name='/lfs1/credentials/key.der' len=121 - file_management:133
../../certs/playground-2f69a2-0000000001.key.der ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0.0% • 0/121 bytes • ? • -:--:--
```

With a manually specified mtu:
```
(golioth) samfriedman ~/golioth/bluetooth-gateway/bluetooth-gateway [main] $ smpmgr --port /dev/cu.usbmodem0010502274171 --mtu 128 file upload ../../certs/playground-2f69a2-0000000001.crt.der /lfs1/credentials/crt.der
⠋ Connecting to /dev/cu.usbmodem0010502274171... OK
../../certs/playground-2f69a2-0000000001.crt.der ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.0% • 352/352 bytes • 1.0 kB/s • 0:00:00
```

JPHutchins: Years ago, it was possible to compile with incompatible configs. There was some investigation [here](https://github.com/intercreate/smpclient/issues/22#issuecomment-2123736901), and in smpclient's integration tests: 

If you're using USB and can spare the RAM, I recommend a single buffer of 1024 for a nice balance of speed and size.

PR #41, opened 2025-07-22
Use updated info characteristic

trond-snekvik:
Changes the format of the info characteristic to flags and server cert serial number. This allows us to remove the init function for the attributes, as we have to construct the info structure before each read, in case any of the data changes. This simplifies the macro, setup and linker sections.

Renames the peripheral_create to init to reflect the lack of a destroy function, now that we don't have to allocate anything.


sam-golioth (approved):
Much simpler! Nice work @trond-snekvik

hasheddan (approved):

PR #46, opened 2025-07-23
transport: ble_gatt: info: fix build for plaintext encryption

sam-golioth:
When Pouch is compiled without encryption support, the info characteristic failed to link due to an unresolved reference to `pouch_server_certificate_serial_get()`.

### DIFF
--- a/src/transport/ble_gatt/info_characteristic.c
+++ b/src/transport/ble_gatt/info_characteristic.c
@@ -39,12 +39,17 @@ static struct info_ctx
 
 static int build_info_data(void)
 {
+#if CONFIG_POUCH_ENCRYPTION_SAEAD
     uint8_t snr[CERT_SERIAL_MAXLEN];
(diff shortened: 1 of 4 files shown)


hasheddan (approved):
@sam-golioth can we add this build to CI to ensure we catch this in the future?

src/transport/ble_gatt/info_characteristic.c
```diff
@@ -39,12 +39,17 @@ static struct info_ctx
 
 static int build_info_data(void)
 {
+#if CONFIG_POUCH_ENCRYPTION_SAEAD
     uint8_t snr[CERT_SERIAL_MAXLEN];
     ssize_t snr_len = pouch_server_certificate_serial_get(snr, sizeof(snr));
```
mniestroj, line 44: How about providing
```
static inline pouch_server_certificate_serial_get(uint8_t +snr, size_t len)
{
    snr = NULL;
    return 0;
}
```
in the header file when using plaintext?

PR #50, opened 2025-07-25
golioth_sdk: prefix log modules with glth_

hasheddan:
Prefixes all golioth_sdk log modules with `glth_` to avoid conflicts with other log modules, such as the zephyr settings subsystem. `glth_` is used instead of `golioth_` to avoid any conflicts with the existing golioth firmware SDK.


golioth_sdk/dispatch.c
```diff
@@ -5,7 +5,7 @@
  */
 
 #include <zephyr/logging/log.h>
-LOG_MODULE_REGISTER(dispatch, LOG_LEVEL_DBG);
+LOG_MODULE_REGISTER(glth_dispatch, LOG_LEVEL_DBG);
```
hasheddan, line 8: Using `glth_` instead of `golioth_` may be too defensive, given that I don't think we currently envision a scenario where the `golioth_sdk` portion of this repository is used alongside the existing Golioth Firmware SDK. Using the short-hand may also just be nice across the board though. Open to discussion!

sam-golioth (approved):

PR #52, opened 2025-07-29
west: drop import 'path-prefix' (deps)

mniestroj:
Having consisent Zephyr application tree structure allows to quickly
switch between repositories (e.g. to/from bluetooth-gateway), while
having just a single clone of each shared dependency.

This change makes it consistent with golioth-firmware-sdk and
bluetooth-firmware repos and allows to share single west workspace and
switch by invoking 'west config manifest.path MANIFEST_REPO'.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install pip packages
         run: |
           uv pip install \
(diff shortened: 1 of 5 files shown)


hasheddan (approved):

sam-golioth (approved):
I like the `deps` folder, but I concede the point about consistency :)

PR #54, opened 2025-07-29
examples: ble_gatt: show mbedtls error as negative value

mniestroj:
Printing as %x implicitly converts to unsigned. Negate value before
printing, so that meaningful error code is printed.


examples/ble_gatt/src/credentials.c
```diff
@@ -28,7 +28,7 @@ static psa_key_id_t import_raw_pk(const uint8_t *private_key, size_t size)
     int err = mbedtls_pk_parse_key(&pk, private_key, size, NULL, 0, psa_rng_for_mbedtls, NULL);
     if (err)
     {
-        LOG_ERR("Failed to parse key: %x", err);
+        LOG_ERR("Failed to parse key: -%x", -err);
```
sam-golioth, line 31:
```suggestion
        LOG_ERR("Failed to parse key: -0x%x", -err);
```

PR #74, opened 2025-08-14
Remove hard coded log levels

trond-snekvik:
Several pouch modules had their log level hardcoded in the registration macro. This shouldn't be the default, so this change removes all of those, except the ones in the example.


hasheddan (approved):

hasheddan:
(Discussed offline) Would be good to have a config for default log level across modules in the pouch library.

PR #97, opened 2025-09-02
golioth: ota: size OTA status buffer to fit max version and package name

sam-golioth:


golioth_sdk/ota.c
```diff
@@ -26,6 +26,11 @@ LOG_MODULE_REGISTER(ota, CONFIG_GOLIOTH_LOG_LEVEL);
 
 #define GOLIOTH_OTA_COMPONENT_HASH_HEX_LEN 64
 
+#define OTA_STATUS_FIXED_SIZE 21
+#define OTA_STATUS_ENCODE_BUF_SIZE                                  \
+    (OTA_STATUS_FIXED_SIZE + 2 * CONFIG_GOLIOTH_OTA_MAX_VERSION_LEN \
+     + CONFIG_GOLIOTH_OTA_MAX_PACKAGE_NAME_LEN)
```
hasheddan, line 32: Could be worth noting here that we have 2x the max version length to accommodate for reporting the `version` and the `target` 👍🏻

hasheddan (approved):
Thanks @sam-golioth! One non-blocking comment 👍🏻

PR #104, opened 2025-09-29
Uplink indications

sam-golioth:
Support using indications to send uplink. Explicit reads from the Gateway are still supported.

Resolves golioth/firmware-issue-tracker#845

### DIFF
--- a/src/transport/ble_gatt/golioth_ble_gatt_declarations.h
+++ b/src/transport/ble_gatt/golioth_ble_gatt_declarations.h
@@ -27,6 +27,10 @@
         BT_GATT_ATTRIBUTE(_uuid, _perm, _read, _write, _user_data);                            \
     GOLIOTH_BT_GATT_CUD(name)
 
+#define GOLIOTH_BLE_GATT_CCC(name, _changed, _write, _perm) \
+    STRUCT_SECTION_ITERABLE(bt_gatt_attr, name##_zzz_ccc) = \
+        BT_GATT_CCC_WITH_WRITE_CB(_changed, _write, _perm)
+
 #define GOLIOTH_BLE_GATT_SERVICE(_uuid) \
     STRUCT_SECTION_ITERABLE(bt_gatt_attr, AAA_golioth_ble_gatt_svc) = BT_GATT_PRIMARY_SERVICE(_uuid)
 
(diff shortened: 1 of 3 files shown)


src/transport/ble_gatt/uplink_characteristic.c
```diff
@@ -85,10 +104,151 @@ static ssize_t uplink_read(struct bt_conn *conn,
...
+    if (value & BT_GATT_CCC_INDICATE)
+    {
+        LOG_DBG("Indications enabled");
+
+        send_indication(ctx);
+    }
+    else
+    {
+        LOG_DBG("Indications disabled");
+        /* TODO: Indications turned off, stop sending */
```
hasheddan, line 247: Are we avoiding cleaning up context here because we may want to resume sending without the client explicitly writing to the CCC again?

src/transport/ble_gatt/golioth_ble_gatt_declarations.h
```diff
@@ -27,6 +27,10 @@
         BT_GATT_ATTRIBUTE(_uuid, _perm, _read, _write, _user_data);                            \
     GOLIOTH_BT_GATT_CUD(name)
 
+#define GOLIOTH_BLE_GATT_CCC(name, _changed, _write, _perm) \
+    STRUCT_SECTION_ITERABLE(bt_gatt_attr, name##_zzz_ccc) = \
+        BT_GATT_CCC_WITH_WRITE_CB(_changed, _write, _perm)
```
hasheddan, line 32: Looks like this wasn't [introduced](https://github.com/zephyrproject-rtos/zephyr/commit/7b61bd65187d87641c6644fb19cfce1507b7fd13) until Zephyr v4.2.0 so we'll either need to bump or use the more cumbersome old style.

trond-snekvik (commented):
:tada:

src/transport/ble_gatt/uplink_characteristic.c
sam-golioth: Although, taking another look, we should probably cleanup the context in the changed callback instead of the write callback, because the write callback can be called even if the value hasn't changed (i.e. indications were already disabled).

src/transport/ble_gatt/golioth_ble_gatt_declarations.h
sam-golioth: The update to 4.2 didn't require any changes so I've gone ahead and included it in this PR. Tests are passing now.

hasheddan (approved):

PR #119, opened 2026-01-28
Implement HIL testing

szczys:
Implement Hardware-in-the-Loop testing.

- Separate build from integration tests
- Add gateway build
- Add pytest for Settings service to `example/ble_gatt`

## Known Limitations

- Gateway
  - Currently the `frdm_rw612` using Ethernet is the only supported gateway device for HIL testing
  - While the gateway version may be specified, only one version may be used for any given test run
- HIL build won't start until the Gateway build finishes
  - To avoid building the gateway for every matrix strategy, the HIL build/test job `needs` the Gateway build to finish before the individual HIL build steps may begin. One alternative to this would be to split the HIL build/test steps into their own workflow. That approach was not used as it would have required the matrix stategy to be specified twice (once for each of those workflow calls).
- Initial implementation only supports node testing on NCS
  - HIL tests are run via pytest (no twister). Because of this, nrf52840 binaries built for vanilla Zephyr don't run on the device, likely because both mcuboot and application code needs to be flashed. This issue is [tracked here](https://github.com/golioth/firmware-issue-tracker/issues/908).
- Testing in Dev is currently not supported. This issue is [tracked here](https://github.com/golioth/firmware-issue-tracker/issues/905).

Resolves https://github.com/golioth/firmware-issue-tracker/issues/901
Resolves https://github.com/golioth/firmware-issue-tracker/issues/902

### DIFF
--- a/.github/workflows/build-gateway.yml
+++ b/.github/workflows/build-gateway.yml
@@ -1,0 +1,92 @@
+name: "Build Gateway"
+
+on:
+  workflow_dispatch:
+    inputs:
+      gw_board:
+        required: false
+        type: string
+      gw_version:
+        required: false
+        type: string
+      gw_manifest_yml:
+        required: false
+        type: string
+      gw_blobs:
+        required: false
+        type: string
+
+  workflow_call:
+    inputs:
+      gw_board:
+        required: false
+        type: string
+      gw_version:
+        required: false
+        type: string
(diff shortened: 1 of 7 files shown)


trond-snekvik (approved):

szczys:
This work has fallen very far behind our development and at this point should be completely reimplemented.

PR #134, opened 2026-03-22
DNM: propsed API for the internal port layer

szczys:
This draft PR can be used to discuss the proposed port API and find alignment using the review process, however it should not be merged to main.

Perhaps [the best way to view the proposed API](https://github.com/golioth/pouch/blob/szczys/port-survey/port/proposed-port-api.md) is using the markdown preview.

### DIFF
--- a/port/proposed-port-api.md
+++ b/port/proposed-port-api.md
@@ -10,6 +10,12 @@ Opaque pointers have been avoided throughout so that the API functions
 don't allocate memory (statically or dynamically). The caller controls
 its own memory when instantiating typedefs.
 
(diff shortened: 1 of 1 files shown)


trond-snekvik (commented):
Transport isn't included here, but since we only have the BLE transport, we'll need it to work on the target platforms to be able to properly test the port. I think we should finish restructuring the transport first though, as that should make it a lot easier to swap out the underlying BLE API.

szczys:
This has been implemented with the exception of:
- errno.h: we assume the RTOS provides it and are including it where necessary
- delayable work: This will be needed to add BLE cross-platformability and will be part of future work.

PR #142, opened 2026-03-30
Split BLE GATT transport layer

trond-snekvik:
Moves the BLE GATT transport's sender and receiver to a new directory named "sar" (segmentation and reassembly) that handles the part of toothfairy that can be transport agnostic, namely the segmentation and acking logic.

The BLE GATT specific code has been abstracted under a new, generic "trx" API, representing the raw transport.

The characteristics have been moved to "topics", which are transport agnostic data channels/sessions/topics that handle the glue between the underlying raw transport and the pouch transport APIs.

The naming conventions should be discussed here. I picked what felt the most natural, but upon reviewing the system as a whole, I wonder if something that's closer to the OSI model naming would be better. For instance, network, transport and session might map well as replacements for trx, sar and topics, respectively. "Transport" is being used for the entire thing at the moment though, so I decided to hold off on renaming.

Although these new files are technically moved versions of the existing characteristics, sender, receiver and packetizer, I chose to reinstate the files that are used in the gateway to avoid any significant conflicts with other ongoing work in the porting and gateway efforts. The intent would be to remove the common GATT library and move the gateway to use the same transport as the device once everything is settled.

The refactor is built on the following principles and assumptions:
- The packetizer.c file included both the packet format and half of the packet splitting functionality, which obfuscated the control flow unnecessarily. The first move was to split this file, so that the packet serialization and deserialization was isolated from the control path.
- The architecture split for the packetizer, sender/receiver and characteristics modules have been muddied over time, which made the packets flow through each module in paths that were hard to follow and untangle. Together, the three parts essentially did four things (but each task had kind of spilled out across the three modules):

  - Parse packets
  - Flow control
  - Characteristic specific data fetching/forwarding
  - BLE GATT specific rx/tx

  By splitting the modules accordingly (packet.c, sender.c/receiver.c,
  topics and peripheral.c, respectively), the overall code is easier to
  follow, and feel more like a layered architecture.
- The GATT characteristics won't really be used separately, and since they all use the packetizer, they're all either senders or receivers that work the same way. The only difference between them is what API they send data to and from, so they have more in common with each other than they do with their owners. Therefore, it's cleaner to gather them all in the same file and use the same abstraction for all of them than it is to create a distributed macro system. This will be true for any transport that needs segmented messaging.
- We'll want to test the segmentation and reassembly mechanism in isolation, and by creating a common abstraction for the underlying transport (trx.h) we can create test transports. By allowing the topics to instantiate the senders and receivers, we get a testable API at the top as well.
- The window size is a property the transport layer, not the topics (info, downlink), and should be configured accordingly.

Overview of moved/renamed files:

image

Overall architecture:

image


szczys (approved):
This is some beautiful work @trond-snekvik, well done!

I don't have strong feelings on the naming scheme. It's not obvious that `sar` is segmentation and reassembly and `trx` is transport. I want to suggest we document that somewhere... but I don't have advice on where to do that so maybe it's not needed?

trond-snekvik:
@szczys I followed your suggestion, and gave a description of the system to a clanker without using any of the names from this PR :robot:  It suggested:
- "Endpoints" instead of "topics"
- "SAR Engine" for the segmentation and reassembly/acking part - pretty much the same as in the PR
- "bearer" for the underlying medium, ie trx.h

I kinda like these suggestions better than the original names - endpoints are pretty appropriate in the transport context, and bearer works well for me (e.g. BLE GATT Bearer, SPI bearer). Thoughts?

src/transport/topics/server_cert.c
```diff
@@ -0,0 +1,70 @@
...
+        struct pouch_cert server_cert = {
+            .buffer = cert.buffer,
+            .size = cert.size,
+        };
+        pouch_server_certificate_set(&server_cert);
+    }
+
+    free(cert.buffer);
+    cert.buffer = NULL;
+}
```
hasheddan: I'm curious if we can use the topics (maybe soon to be known as endpoints) with transports that do not use SAR? For example, we have essentially the same code in [the HTTP client](https://github.com/golioth/pouch/blob/32e1735b266d185058534f16cd2289ccaaae3e40/port/zephyr/transport/http/client.c#L122). I think it _could_ make implementing transports simpler / more uniform. I'm not necessarily concerned about it being done in this PR for the HTTP client, but I think it could impact how we organize the code (and to the point of my other comment, whether we pass `pouch_receiver` / `pouch_sender`).

mniestroj (approved):

hasheddan (commented):
Awesome work @trond-snekvik! I like "SAR" and "endpoints" (kind of fun how SAR matches both "segmentation and reassembly" as well as "sender and receiver" 😉 ). I don't love bearer because of my own association of bearer with a bearer token. However, based on my comments below, I wonder if we may actually end up being able to have the trx / bearer layer just be the transport layer (if every transport sits at the same level in the hierarchy).

You have certainly been more hands on here, so I expect there are some good reasons not to do what I describe in my comments. Hopefully it will at least help us enumerate those reasons. Happy to do a quick white-boarding session as well!

hasheddan (commented):
Awesome stuff @trond-snekvik! LGTM, though left a comment below about some of the weirdness that will need to be fixed in rebase on `main` :/

hasheddan (approved):
LGTM, nice work @trond-snekvik!

PR #160, opened 2026-04-15
Port: cleanup non-portable code preventing ESP-IDF platform builds

szczys:
This PR makes changes throughout the tree to facilitate building on ESP-IDF. Highlights include:

- Logging
  - Using port logging calls
  - Explicitly defining Kconfig log levels in place of Zephyr-centered Kconfig includes
  - Casting variables to resolve type errors (`uint8_t *` vs. `char *`, etc)
- Remove erroneous Zephyr includes which are no longer needed but remained in some files
- Define `IS_ENABLED()` when not already defined
- Add explicit `#include <errno.h>` to files using posix error codes

### DIFF
--- a/golioth_sdk/ota_upper.c
+++ b/golioth_sdk/ota_upper.c
@@ -7,8 +7,6 @@
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(ota_upper, CONFIG_GOLIOTH_LOG_LEVEL);
 
-#include <zephyr/kernel.h>
-
 #include <pouch/golioth/ota.h>
 #include <pouch/port.h>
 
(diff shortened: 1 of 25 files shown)


golioth_sdk/ota_upper.c
```diff
@@ -7,8 +7,6 @@
 #include <zephyr/logging/log.h>
 LOG_MODULE_REGISTER(ota_upper, CONFIG_GOLIOTH_LOG_LEVEL);
```
hasheddan: @szczys this also needs to be switched to `POUCH_LOG_REGISTER` I believe.

szczys: Thanks for catching, I've added another commit!

hasheddan (approved):
Thanks @szczys!

PR #162, opened 2026-04-15
Port: include server CA cert as a header file

szczys:
Make the server CA cert available as a `pouch_cert` struct via header inclusion.

### DIFF
--- a/port/esp_idf/pouch_ca_cert.c
+++ b/port/esp_idf/pouch_ca_cert.c
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2026 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <pouch/pouch_ca_cert.h>
+#include <pouch/types.h>
+#include <pouch/port.h>
+
+extern const uint8_t pouch_ca_cert_der_start[] asm("_binary_pouch_ca_cert_der_start");
+extern const uint8_t pouch_ca_cert_der_end[] asm("_binary_pouch_ca_cert_der_end");
+
+struct pouch_cert raw_ca_cert;
+
+static void populate_ca_cert_struct(void)
+{
+    raw_ca_cert.buffer = pouch_ca_cert_der_start;
+    raw_ca_cert.size = pouch_ca_cert_der_end - pouch_ca_cert_der_start;
+};
+POUCH_APPLICATION_STARTUP_HOOK(populate_ca_cert_struct);
+
+struct pouch_cert *pouch_ca_cert = &raw_ca_cert;
--- a/port/include/pouch/pouch_ca_cert.h
+++ b/port/include/pouch/pouch_ca_cert.h
@@ -1,0 +1,11 @@
+/*
+ * Copyright (c) 2026 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <pouch/types.h>
+
+extern struct pouch_cert *pouch_ca_cert;
--- a/port/zephyr/CMakeLists.txt
+++ b/port/zephyr/CMakeLists.txt
@@ -25,7 +25,9 @@ if(CONFIG_POUCH)
 
         generate_inc_file_for_target(pouch
             ${ca_cert}
-            ${ZEPHYR_BINARY_DIR}/include/generated/golioth_ca_cert.inc)
+            ${ZEPHYR_BINARY_DIR}/include/generated/pouch_ca_cert.inc)
+
+        zephyr_library_sources(pouch_ca_cert.c)
     endif()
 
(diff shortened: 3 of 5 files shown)


port/zephyr/golioth_ca_cert.c
```diff
@@ -0,0 +1,19 @@
...
+#include <pouch/types.h>
+
+static const uint8_t golioth_ca_cert_der[] = {
+#include "golioth_ca_cert.inc"
+};
+
+struct pouch_cert golioth_ca_cert = {
+    .buffer = golioth_ca_cert_der,
+    .size = sizeof(golioth_ca_cert_der),
+};
```
hasheddan: I don't think we should have any reference to `golioth` in the porting layer, these could just as easily be `pouch_ca_cert_der`, `pouch_ca_cert.inc`, `pouch_ca_cert.c`, etc.

I'm also thinking it may make more sense to, rather than expanding the porting layer to support providing a CA certificate for `pouch`, just expand [`pouch_config`](https://github.com/golioth/pouch/blob/ae368b8e99ede9560b22e316999becd90c0cd011/include/pouch/pouch.h#L14), then handle loading the certificate for the specific platform in the examples. Someone may want to provide a CA certificate in some manner other than compiling it into the firmware image, and we should allow them to do so without having to modify the library / porting layer. Expanding `pouch_config` means that the application can control how a CA certificate is obtained.

szczys: > I don't think we should have any reference to golioth in the porting layer, these could just as easily be pouch_ca_cert_der, pouch_ca_cert.inc, pouch_ca_cert.c, etc.

Good callout. I've removed the mention of Golioth in the filenames and variables.

> I'm also thinking it may make more sense to, rather than expanding the porting layer to support providing a CA certificate for pouch, just expand [pouch_config](https://github.com/golioth/pouch/blob/ae368b8e99ede9560b22e316999becd90c0cd011/include/pouch/pouch.h#L14), then handle loading the certificate for the specific platform in the examples. Someone may want to provide a CA certificate in some manner other than compiling it into the firmware image, and we should allow them to do so without having to modify the library / porting layer. Expanding pouch_config means that the application can control how a CA certificate is obtained.

Currently this CA cert is selected using `CONFIG_POUCH_CA_CERT_FILENAME`. I think requiring it be part of the config is not a huge improvement over that. But I'm not against it. I think for this PR we should not change the current behavior. If you'd like to see this change implemented would you please open an issue for it and assign me?

port/esp_idf/pouch_ca_cert.c
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <pouch/pouch_ca_cert.h>
+#include <pouch/types.h>
+#include <pouch/port.h>
+
+extern const uint8_t pouch_ca_cert_der_start[] asm("_binary_pouch_ca_cert_der_start");
+extern const uint8_t pouch_ca_cert_der_end[] asm("_binary_pouch_ca_cert_der_end");
```
mniestroj, line 12: What makes those `_binary_pouch_ca_cert_der_start` and `_binary_pouch_ca_cert_der_end`? Where do we point ESP-IDF build system to create those symbols? What is the mechanism. I don't see it in this PR.

mniestroj (approved):

port/esp_idf/pouch_ca_cert.c
szczys: You're right that it is not it in this PR. In an upcoming PR (#165) we add Pouch as an ESP-IDF component. In that PR we first find the CA cert based on the Kconfig symbol and copy it into the build directory:

https://github.com/golioth/pouch/blob/0534158c0c410c20f6936060011ffb2e9d408f57/port/esp_idf/components/pouch/CMakeLists.txt#L199-L214

Then use ESP-IDF's `EMBED_FILES` directive to add the binary and generate the symbols you are asking about:

https://github.com/golioth/pouch/blob/0534158c0c410c20f6936060011ffb2e9d408f57/port/esp_idf/components/pouch/CMakeLists.txt#L233-L234

port/zephyr/golioth_ca_cert.c
hasheddan: > I think requiring it be part of the config is not a huge improvement over that.

I think the flexibility it offers quite a bit more flexibility without imposing much additional overhead (i.e. the application can easily support a Kconfig filename, generate the header, compile it in, etc.), but agree we don't need to block on it 👍🏻

hasheddan (approved):

port/zephyr/golioth_ca_cert.c
szczys: Sounds good, I'm tracking this task here: https://github.com/golioth/firmware-issue-tracker/issues/962

hasheddan: Awesome, thanks @szczys!

PR #164, opened 2026-04-15
Port: Zephyr: split common Kconfig into src directory

szczys:
Split pouch common and saead common into Kconfig files that can be included from the port-specific Kconfig "root" files.

### DIFF
--- a/src/Kconfig
+++ b/src/Kconfig
@@ -1,0 +1,58 @@
+# Copyright (c) 2026 Golioth, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+config POUCH_BLOCK_SIZE
+  int "Block size"
+  default 512
+  help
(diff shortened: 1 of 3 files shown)


src/Kconfig
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) 2026 Golioth, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+config POUCH_LOG_LEVEL
+  int "Default log level"
+  default 3
+  help
+    The maximum log level used by pouch module
```
mniestroj: Is this log level integer really portable?

szczys: Thank you for catching this. I added #167 to address this. This PR is rebased to depend on that one.

trond-snekvik (approved):
LGTM, aside from the log part

PR #167, opened 2026-04-16
Port: rework log levels

szczys:
As mentioned by @mniestroj in [a comment on another PR](https://github.com/golioth/pouch/pull/164#discussion_r3093299777), using magic numbers for default log levels is not portable.

Up until now we've used `#define` in port header files to set the logging levels to use for each port. However, these values are not available in Kconfig. To resolve the issue, I've moved the port-level defines to a Kconfig file. The resulting Kconfig symbols are used in `port.h` to set the port layer log level values.

With this approach, we have access to a valid port logging level in both Kconfig and C (without the `CONFIG_` prefix which would have implied that these are not constant). The downside of this is that we're no longer directly mapping to the platform `#define` values, but instead translating those value by hand in the `Kconfig.port_log_level` file. However, log levels are extremely unlike to change upstream so I feel this is a reasonable tradeoff.

As part of this PR, I also adapted the Zephyr logging template Kconfig file to work with our port logging. I had previously removed use of the template, but it is quite handy for adding user-friendly logging level choice menus.


src/Kconfig.template.pouch_log_config
```diff
@@ -0,0 +1,48 @@
...
+	bool "Warning"
+
+config $(module)_LOG_LEVEL_INF
+	bool "Info"
+
+config $(module)_LOG_LEVEL_DBG
+	bool "Debug"
+
+config $(module)_LOG_LEVEL_VERBOSE
+	bool "Debug"
```
hasheddan:
```suggestion
	bool "Verbose"
```

src/Kconfig.template.pouch_log_config
```diff
@@ -0,0 +1,48 @@
...
+
+endchoice
+
+config $(module)_LOG_LEVEL
+	int
+	default POUCH_LOG_LEVEL_NONE if $(module)_LOG_LEVEL_OFF
+	default POUCH_LOG_LEVEL_ERR if $(module)_LOG_LEVEL_ERR
+	default POUCH_LOG_LEVEL_WRN if $(module)_LOG_LEVEL_WRN
+	default POUCH_LOG_LEVEL_INF if $(module)_LOG_LEVEL_INF
+	default POUCH_LOG_LEVEL_NONE if $(module)_LOG_LEVEL_DBG
```
hasheddan:
```suggestion
	default POUCH_LOG_LEVEL_DBG if $(module)_LOG_LEVEL_DBG
```

golioth_sdk/Kconfig
```diff
@@ -52,8 +52,8 @@ config GOLIOTH_OTA_MAX_VERSION_LEN
 
 endif
 
-config GOLIOTH_LOG_LEVEL
-  int "Maximum log level"
-  default 4
+module = GOLIOTH
+module-str = Golioth
+rsource "../src/Kconfig.template.pouch_log_config"
```
hasheddan, line 57: I suspect we will no longer do this when the Golioth SDK is portable as the template is Zephyr specific. Would it make sense for the template to live in the Zephyr port instead?

src/Kconfig.template.pouch_log_config
szczys: Oof, that is a particularly egregious error. Thank you for catching!

golioth_sdk/Kconfig
szczys: This file is needed cross platform. I placed it in `src` because in #164 this section of code gets moved to `src/Kconfig`. I also think this template is the approach we should use whenever adding a new log level group to pouch. I don't know how often that will happen but any cross-platfor libs, sdks, transports, should probably have their own configurable level.

mniestroj (approved):
LGTM

hasheddan (approved):
Thanks @szczys!

PR #173, opened 2026-04-21
Port: ESP-IDF: Add Uplink

szczys:
- Add Kconfig for http_transport
- Add flags to protect access and limit server cert download and pouch cert upload
- Send device cert to server
- Send uplink to server


port/esp_idf/test/http_client/sdkconfig.defaults
```diff
@@ -1 +1,5 @@
+CONFIG_ESP_MAIN_TASK_STACK_SIZE=5120
+CONFIG_POUCH_THREAD_STACK_SIZE=4096
+CONFIG_POUCH_UPLINK_PROCESSING_STACK_SIZE=2560
```
szczys: I tested these a bit just now and the first two are quite oversized for just the uplink. At some point I was seeing stack overflows which led me to grow these sizes. Once we have downlink running I will test again and tune them to be closer to what is actually being used.

trond-snekvik (approved):

port/esp_idf/test/http_client/main/http_client.c
```diff
@@ -62,77 +77,383 @@ static esp_err_t pouch_server_cert_response_callback(esp_http_client_event_t *ev
...
+        .cert_len = sync->mtls_creds->cert_len,
+        .client_cert_pem = sync->mtls_creds->client_cert_pem,
+        .client_cert_len = sync->mtls_creds->client_cert_len,
+        .client_key_pem = sync->mtls_creds->client_key_pem,
+        .client_key_len = sync->mtls_creds->client_key_len,
     };
 
-    /* TODO: replace log with call to pouch_server_certificate_set(&cert); */
-    ESP_LOG_BUFFER_HEXDUMP(TAG, cert.buffer, cert.size, ESP_LOG_WARN);
+    esp_http_client_handle_t client = esp_http_client_init(&config);
```
hasheddan: Rather than initializing and cleaning up the client on every operation, we should try to use the same client handle and use the `esp_http_client_set_*` functions. This will allow us to reuse connections, whereas the current strategy will result in performing a new handshake on every single request.

szczys: Thanks for the call out on this this. I added a commit to implement the reusable http client during the server cert download and updated the pouch cert upload and uplink commits to use it.

port/esp_idf/test/http_client/main/main.c
```diff
@@ -55,5 +69,12 @@ void app_main(void)
...
     ESP_LOGI(TAG, "Pouch successfully initialized");
 
+    while (true)
+    {
+        /* Sync Pouch uplink and downlink */
+        http_client_transport_sync();
+
+        vTaskDelay(pdMS_TO_TICKS(CONFIG_EXAMPLE_HTTP_CLIENT_SYNC_PERIOD_S * 1000));
+    }
     http_client_transport_sync();
```
szczys:
```suggestion
```

port/esp_idf/test/http_client/main/http_client.c
```diff
@@ -62,77 +101,408 @@ static esp_err_t pouch_server_cert_response_callback(esp_http_client_event_t *ev
...
+        if (POUCH_NO_MORE_DATA == res)
+        {
+            *is_last = true;
+            break;
+        }
+
+        if (available_space)
+        {
+            vTaskDelay(pdMS_TO_TICKS(100));
+        }
```
hasheddan: Why do we need to delay here?

port/esp_idf/test/http_client/main/http_client.c
```diff
@@ -62,77 +101,408 @@ static esp_err_t pouch_server_cert_response_callback(esp_http_client_event_t *ev
...
+static int send_pouch_uplink(struct sync_context *sync)
+{
+    sync->uplink = pouch_uplink_start();
+    if (NULL == sync->uplink)
+    {
+        ESP_LOGE(TAG, "Failed to start uplink");
+        return -ENOMEM;
+    }
+
+    sync->server_cert.pos = 0;
```
szczys:
```suggestion
```

port/esp_idf/test/http_client/main/http_client.c
szczys: That is a great question to which I cannot remember the answer. This copied from the Zephyr http client transfer. I vaguely remember having some issues with a tight loop here when pouch indicated there was more data to be copied. But I don't recall why time needs to be taken (instead of `yield()` but I don't see why that would be necessary either)

port/esp_idf/test/http_client/main/http_client.c
```diff
@@ -62,77 +101,408 @@ static esp_err_t pouch_server_cert_response_callback(esp_http_client_event_t *ev
...
+    {
+        ESP_LOGE(TAG, "Failed to fetch headers: %d", err);
+        return err;
+    }
+
+    int status_code = esp_http_client_get_status_code(sync->client);
+
+    if (!IN_RANGE(status_code, 200, 299))
+    {
+        ESP_LOGE(TAG, "Uplink failed: %d", status_code);
```
szczys, line 356:
```suggestion
        ESP_LOGE(TAG, "Uplink failed: %d", status_code);
        return -EIO;
```

port/esp_idf/test/http_client/main/http_client.c
hasheddan: Hmm can you see if you can reproduce the issue (with Zephyr and here)? I don't love the idea of having delays papering over what may be an underlying issue.

szczys: I tested removing the delay in Zephyr and uplink fails when doing so. However, I'm glad you have drawn attention to this as we should be able to greatly simplify how the uplink is filled. 

I've opened [an issue to work on that in the Zephyr transport](https://github.com/golioth/firmware-issue-tracker/issues/964). As for this transport, I have already simplified the process and pushed an update.

Also of note with this change:
- Removed the bool variable for downlink_started and used an atomic flag
- Reworked how `pouch_uplink_finish()` is called so we don't have any paths that return without finishing

hasheddan (approved):
Thanks @szczys!

PR #174, opened 2026-04-23
examples: http_client: add OTA download test

mniestroj:


examples/zephyr/http_client/src/fw_update_test.c
hasheddan, line 1: This is fairly similar to the flash writing alternative -- thoughts on just making a single implementation, always computing the hash, and just making the config control whether we actually write to flash and reboot or not?

mniestroj: My thoughts were:
- PRO: less maintenance, due to single file to manage
- PRO: adding sha256sum calculation provides some value to the application (we could verify before requesting upgrade to mcuboot) 
- CON: calculating sha256 increases code complexity for end user
- CON: making flash procedure optional further increases complexity for end user

I decided to value end user complexity (example readability) more. I do not have strong feeling about it, so I am definitely open to changes!

hasheddan: @mniestroj that sounds like sound reasoning -- thanks for enumerating! Potentially worth reconsidering if we eventually separate Golioth SDK functionality into dedicated examples, but I think this makes sense for now 👍🏻

hasheddan (approved):
Thanks @mniestroj!

PR #176, opened 2026-04-23
SAR sender: Make use of full transmit window

trond-snekvik:
The SAR sender stops sending when the next sequence number matches the stored window seq, but we're not accounting for that extra number when storing the window from the ack.

Add one, to make use of the full reported window.


hasheddan (approved):
Nice catch @trond-snekvik!

szczys (approved):

PR #178, opened 2026-04-23
port: esp_idf: initialize block after alloc

szczys:
After allocation, blocks must be initialized so that buffer members start at known values.

Resolves: https://github.com/golioth/firmware-issue-tracker/issues/963


hasheddan (approved):
Nice work chasing this one down @szczys!

PR #179, opened 2026-04-23
Port: ESP-IDF: Add Downlink

szczys:
- Update golioth_sdk files for portability
  - add weakly-defined hex2bin for portability
  - use pouch logging
  - replace ARRAYSIZE() calls with portable calculation
  - include necessary header files
- Build golioth_sdk as an ESP-IDF module
  - [future work](https://github.com/golioth/firmware-issue-tracker/issues/967) will do the same for Zephyr
- Add downlink to the transport and example application


szczys:
This needs a couple of changes so the chunked get works properly
- Call `perform` after the chunk uplink instead of calling `close`
- Use the `FINISH` event to close a chunked get

These are fairly small changes, but I'm going to set it to draft for now and will update in the morning.

trond-snekvik (approved):

hasheddan (approved):
Nice work @szczys!

PR #181, opened 2026-04-24
Gateway: Split transport

trond-snekvik:
Splits the Gateway transport layer using the same modules as the pouch device transport.

Removes the BLE GATT Common lib, including the packetizer and old sender/receiver. Moves the BLE adv data definition to include/pouch/transport/ble/gatt.h.

The application facing gateway Bluetooth API was partially implemented in the main file. This has been changed to use the callback mechanism we're using elsewhere, allowing the application to decide what to do when the gateway operation is done. To make this consistent and flexible, the gateway itself no longer performs a disconnect, it tells the app that it's done with its gateway operation for the connected device.

The BLE part of the gateway transport machinery is now limited to gateway.c, which sets up bearers for each characteristic. This also handles all processing through the transport state machine, instead of having each action decide what happens next when it completes.

The bearer pointer now has a context pointer, and is passed up to the endpoint callbacks. This allows the gateway to determine which device the current endpoint operation is targeting, and provides a way for endpoints to interact with the bearer, for instance by closing them on asynchronous errors.

The bearer interface got three new additions:
- The bearer can now be nudged with a ready() function, allowing us to resume the bearer operation when new data comes in.
- The bearer abort function has been replaced by bearer_close(), which takes a success flag. This allows the gateway bearer to process the end of each transaction. Previously, we only told the endpoints about successful transactions.
- The window member of the bearer has been removed. This was only used by the receiver, which now keeps track of this itself.

The transport/endpoints directory has been split into transport/endpoints/device and transport/endpoints/gateway to keep them separated.


src/transport/endpoints/gateway/endpoints.h
```diff
@@ -0,0 +1,14 @@
...
+ */
+#pragma once
+
+#include "../endpoint.h"
+
+extern const struct pouch_endpoint gateway_endpoint_info;
+extern const struct pouch_endpoint gateway_endpoint_device_cert;
+extern const struct pouch_endpoint gateway_endpoint_server_cert;
+extern const struct pouch_endpoint gateway_endpoint_uplink;
+extern const struct pouch_endpoint gateway_endpoint_downlink;
```
hasheddan: @trond-snekvik this might be something we have to incrementally move towards, but I am envisioning there being `transport/endpoints/server` and `transport/endpoints/device` rather than `device` and `gateway`. It should be (for the most part) inconsequential whether the "pouch device' here is also acting as a "pouch server" to the cloud (gateway) or not (local processing).

```suggestion
extern const struct pouch_endpoint server_endpoint_info;
extern const struct pouch_endpoint server_endpoint_device_cert;
extern const struct pouch_endpoint server_endpoint_server_cert;
extern const struct pouch_endpoint server_endpoint_uplink;
extern const struct pouch_endpoint server_endpoint_downlink;
```

I _think_ that we could go ahead and move to this structure as you have already effectively decoupled the BLE interaction (device <-> gateway) from the HTTP interaction (gateway <-> cloud). Then the next step would be removing the `golioth_client` and instead using an HTTP / CoAP "client transport" as a "pouch server" for gateway <-> cloud. What do you think?

include/pouch/transport/bluetooth/gatt.h
hasheddan, line 66: If the `include/pouch/gateway/bt` portion is moved to this directory as `gatt_client`, then this could make sense to move into a `gatt_server` header as it is only relevant for the GATT server whereas the rest of this file is relevant for both.

include/pouch/gateway/bt/connect.h
```diff
@@ -8,33 +8,13 @@
 
 #include <zephyr/bluetooth/conn.h>
 
-#include <pouch/gateway/types.h>
+/** Bluetooth gateway end callback */
+typedef void (*pouch_gateway_bt_end_t)(struct bt_conn *conn);
```
hasheddan, line 12: This is not introduced in this PR, but we shouldn't be passing a Zephyr-specific `bt_conn` in the user facing ~gateway~ server API. I believe the remainder of the `include/pouch/gateway/bt` directory (scan and bond) is not Zephyr-specific (other than some `k_timeout_t` references that should be able to be replaced fairly easily), and, related to my other comment, I think these headers could be moved to the `include/transport/bluetooth` directory. It would add scope to this PR though, so likely makes more sense to do as a subsequent effort 👍🏻

src/transport/endpoints/gateway/downlink.c
```diff
@@ -0,0 +1,63 @@
...
+
+#include <errno.h>
+#include <pouch/gateway/downlink.h>
+#include "gateway/types.h"
+#include "transport/bearer.h"
+#include "transport/bearer.h"
+#include "endpoints.h"
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(downlink_endpoint, CONFIG_POUCH_GATEWAY_LOG_LEVEL);
```
hasheddan: Any reason not to use portable logging here?

src/transport/endpoints/gateway/server_cert.c
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#include <errno.h>
+#include <pouch/gateway/cert.h>
+#include "gateway/types.h"
+#include "transport/bearer.h"
+#include "endpoints.h"
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(server_cert_endpoint, CONFIG_POUCH_GATEWAY_LOG_LEVEL);
```
hasheddan: Same note on portable logging.

src/transport/endpoints/gateway/uplink.c
```diff
@@ -0,0 +1,65 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <pouch/gateway/uplink.h>
+#include "gateway/types.h"
+#include "transport/bearer.h"
+#include "endpoints.h"
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(uplink_endpoint, CONFIG_POUCH_GATEWAY_LOG_LEVEL);
```
hasheddan: Same note on portable logging.

src/transport/endpoints/device/endpoints.h
```diff
@@ -0,0 +1,14 @@
...
+ */
+#pragma once
+
+#include "../endpoint.h"
+
+extern const struct pouch_endpoint pouch_endpoint_info;
+extern const struct pouch_endpoint pouch_endpoint_device_cert;
+extern const struct pouch_endpoint pouch_endpoint_server_cert;
+extern const struct pouch_endpoint pouch_endpoint_uplink;
+extern const struct pouch_endpoint pouch_endpoint_downlink;
```
hasheddan: In correspondence with my comment about changing "gateway" to "server":
```suggestion
extern const struct pouch_endpoint pouch_device_endpoint_info;
extern const struct pouch_endpoint pouch_device_endpoint_device_cert;
extern const struct pouch_endpoint pouch_device_endpoint_server_cert;
extern const struct pouch_endpoint pouch_device_endpoint_uplink;
extern const struct pouch_endpoint pouch_device_endpoint_downlink;
```

hasheddan (commented):
Nice work navigating the complexity here @trond-snekvik! Most of my comments below are related to moving towards a future desired structure, so it is likely that they may be deferred, though I think some of them are actionable here. However, I do want to ensure we are aligning on the same terminology and concepts. I am envisioning the following:

**Roles**
- **Pouch Device** (one-to-one communication)
  - _Produces_ pouches for a _single_ entity (a pouch server)
  - _Consumes_ pouches from a _single_ entity (a pouch server)
- **Pouch Server** (one-to-many communication)
  - _Produces_ pouches for _multiple_ entities (pouch devices)
  - _ Consumes_ pouches from _multiple_ entities (pouch devices)

Importantly, a gateway doesn't "exist" in the pouch framework. A device always talks to a server, which may happen to be a gateway. A server always talks to a device, which may happen to be a gateway. A gateway is simply an entity that happens to act as both a server (to other devices, including potentially itself), and a device (to a server).

Image


A single entity could dynamically transition between acting as a gateway and acting as a server (i.e. switch from forwarding pouches to handling locally), but that should not be relevant  from the perspective of any pouch device or pouch server.

Following on from this foundation, we have the following types of endpoints:

**Endpoints**
- **Device**
  - _Produces_ **device_cert**
  - _Consumes_ **server_cert**
  - _Produces_ **uplink**
  - _Consumes_ **downlink**
  - _Produces_ **info** (optional)
- **Server**
  - _Consumes_ **device_cert**
  - _Produces_ **server_cert**
  - _Consumes_ **uplink**
  - _Produces_ **downlink**
  - _Consumes_ **info** (optional)
 
Importantly, there are no "gateway" endpoints because endpoints are fundamentally exposing functionality to another entity, and a "gateway" is not a role that any pouch entity is knowledgeable of.

Transports may be the most confusing part because they offer their own concepts of clients and servers (which is why we explicitly avoid using the term "client" when talking about pouch entities -- perhaps worth considering if using a term other than "server" would help make things clearer). Device and server endpoints can be implemented by both transport clients and servers, which may influence the interaction model. The _optional_ **info** endpoint is the clearest evidence of this attribute, as if the Pouch Server is a _transport client_ (e.g. GATT client), it must request information to know when to consume from the Pouch Device. If the Pouch Server is a _transport server_, the Pouch Device is a _transport client_ and can produce data for the server at its discretion.

Though many of them may not be relevant in feasible use cases, I think it is helpful to think through how any transport we are implementing could act as both a device and server (e.g. swap around the roles that GATT client and server are being used for).

hasheddan (commented):
Following up here with summary of team discussion today:

Generally aligned on their being only two roles (referred to as Device and Server in previous comments). Defining only two roles ensures that we are able to support arbitrary future network topology, as there is no strict definition of a "gateway". However, "server" is easily conflated with a transport concept of a server in a client / server models. Other terminology, such as actor / director, is not suitable because it implies that one entity is always initiating communication, which is not the case with pouch.

The concept of a Pouch Server is most similar to something like an MQTT broker, but it is transport agnostic. The term Pouch Broker fits well with an N-hop topology because intermediary entities (i.e. "gateways") are truly acting as a broker, as is the final destination. The difference is that the intermediary entity is non-terminating, whereas the final destination entity is terminating. 

We'll switch from using Pouch Device and Pouch Server, to Pouch Device and Pouch Broker. An intermediary entity may transition from acting as non-terminating to terminating, but doing so is not evident to Devices, which continue to see it as just a Broker. In this PR, we'll move the gateway endpoints -> ~server~ broker endpoints. Any change to the use of the [server pouch role in session IDs](https://github.com/golioth/pouch/blob/9d571f38de9a26a7b003d38b4c8f9b4b69a50ddb/src/pouch.h#L16) we'll be addressed in follow-up work as it will be a breaking change.

szczys (approved):
Great work @trond-snekvik 🚀

include/pouch/gateway/bt/connect.h
trond-snekvik: I was trying to incorporate this and the above comment, but I wonder if we might want to take a step back here first.

The other APIs in gateway/bt/*.h are about managing connections and security. In the pouch device, we make the application take care of this part - starting advertisements and accepting connection requests. The Pouch device Bluetooth module is just the operation of the GATT server.

This new header is the only part that is actually about the broker functionality (ie the machinery to get pouches up and down to the device) in this directory. 

So the question is which parts of the gateway role belongs to the module, and which parts belong to the application? Also:

- This file is technically the Bluetooth GATT Client broker API - should it go in a different directory, not named gateway?
- Should the connection handling be part of the module or the application (example)?
  - Is there a gateway library separate from the broker library to help users implement scanning and connecting?

I think the last question is essentially the same as the ones we had for pouch devices, and ideally, the answer should be the same:
- The application is responsible for managing connections
- We will make modules that implement the normal usage scenario on top of the core functionality in the future
- For now, we put the connection handling in the application, and see the helper modules as a UX improvement for later

Thoughts? I'd like to do this all separately from this PR, as it's not new here, but renaming this file is probably something we'll want to do in this PR. 

I'm not entirely sure what the best way to handle the platform specific aspects of this API is though. We could make it zephyr specific, and start making some parts of the API specific to each port, or we could make an alternative platform agnostic representation of `struct bt_conn`. I think I would prefer the former, but it could be a bit of a Pandora's box situation

hasheddan: > I think the last question is essentially the same as the ones we had for pouch devices, and ideally, the answer should be the same

Agree with this and the following points 👍🏻 

> I'm not entirely sure what the best way to handle the platform specific aspects of this API is though. We could make it zephyr specific, and start making some parts of the API specific to each port, or we could make an alternative platform agnostic representation of `struct bt_conn`. I think I would prefer the former, but it could be a bit of a Pandora's box situation

I am curious if a modification of the `broker_bt_gatt_device` that is currently in the Zephyr port could be the integration point here. Based on the current surface area of the broker API here (just an entrypoint for start and a callback for end) I think effectively what we are trying to do is allow the application to control if / how a broker decides to handle a candidate device. With that in mind, I'm curious if we could do something like the following:

- Broker application manages when scanning starts (port-specific similar to the existing `ble_peripheral` implementation for devices).
- Broker application is notified when a candidate device appears, and is "handed" a port-specific representation of a device (i.e. `bt_conn`).
- The port provides a "conn to device" function implementation, which allows the application to translate the port-specific representation to the generic representation. The application then operates on the generic representation, which will effectively boil down to assigning handlers for each endpoint.
  - In reality today, this looks the same for every candidate device that the broker encounters and it is abstracted away from the application. However, if we move towards the broker consuming pouch data and then publishing it as a device facing another broker, then this would be where the application ties its broker pouch transport to its device pouch transport (i.e. replacement for the current firmware SDK functionality).
- The application then hands the generic device representation back to the broker through `pouch_broker_bt_start`, which brokers for the device, using the configured endpoints to consume from device, then publish to upstream broker.

I fully acknowledge that this is skipping over some complexity, but generally I think I am advocating for the application facing broker API exposing (1) when to handle a candidate device (i.e. when to invoke broker start) and (2) how to handle a candidate device (i.e. connect downstream broker pouch machinery to upstream device pouch machinery). Does this architecture resonate with you?

All that to say, I also agree that most of this does not belong in this PR 👍🏻

src/transport/sar/sender.c
```diff
@@ -156,55 +155,70 @@ int pouch_sender_recv(struct pouch_sender *sender, const uint8_t *buf, size_t le
...
+{
+    push_fragments(sender);
+}
+
 void pouch_sender_close(struct pouch_sender *sender)
 {
+    if (sender->state == STATE_ACTIVE || sender->state == STATE_READY)
+    {
+        // aborted without reaching the end of the transfer
+        end(sender, false);
```
hasheddan, line 220: Could be worth adding a log (probably at warn level) here.

hasheddan (approved):
One minor comment, but LGTM! Nice work @trond-snekvik!

src/transport/endpoints/uplink.c
```diff
@@ -38,7 +38,7 @@ static void end(bool success)
     uplink = NULL;
 }
 
-const struct pouch_endpoint pouch_endpoint_uplink = {
+const struct pouch_endpoint pouch_device_endpoint_uplink = {
```
mniestroj: This file should be removed, since we define `pouch_device_endpoint_uplink` in `src/transport/endpoints/device/uplink.c`

src/transport/endpoints/broker/downlink.c
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <errno.h>
+#include <pouch/gateway/downlink.h>
+#include <pouch/port.h>
+#include "gateway/types.h"
+#include "transport/bearer.h"
+#include "transport/bearer.h"
```
mniestroj: Double include here.

mniestroj (approved):
LGTM. Two minor issues in comments.

PR #182, opened 2026-04-24
Port: ESP-IDF: Add OTA firmware update to the http client example

szczys:


port/esp_idf/test/http_client/main/fw_update.c
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+GOLIOTH_OTA_COMPONENT(main, "main", CONFIG_APP_PROJECT_VER, ota_main_receive);
+GOLIOTH_OTA_MANIFEST_HANDLER(ota_manifest_receive);
+
+/*
+ * Add a hook used by ESP-IDF CMakeLists.txt to ensure this file is not optimized out
+ */
+void link_hook_fw_update_c(void) {}
```
hasheddan, line 103: I think expecting users to include this workaround in their code is likely to cause quite a bit of issues and confusion. Can we expand the `GOLIOTH_*` macros to create the force link functions on behalf of the user?

szczys: If we create a hook automatically, I don't know how we'd manage the call in user's CMakeLists.txt that matches the symbol:

https://github.com/golioth/pouch/blob/a919859920c937078351a506befa8b293bc786ef/port/esp_idf/test/http_client/main/CMakeLists.txt#L63

We can generate a symbol with the `extern` keyword, but we need to give it a unique name in case the macro is called from different files and we need to call that symbol from somewhere which is tricky if we're generating unique names.

I can think of two ways to remove the need for the hook:
1. Move this code into `main.c` which is already being linked and remove the fw_update.c file.
2. Add an `init` function that must be called; we don't need an init, but it would force the linker to process the file.

hasheddan: @szczys I haven't actually tried this out, but I'm curious if we expanded the `GOLIOTH_*` macros to include a `POUCH_APPLICATION_STARTUP_HOOK` with a no-op function if that would address the issue? It is also fairly hacky, but I suspect it would address the issue without needing to have the user define the function or do any CMake shenanigans.

szczys: Yeah, I have already tried this and it does not cause the file to get processed by the linker:

```c
void init_fw_update(void) {
    printf("FW_UPDATE initialized");
}
POUCH_APPLICATION_STARTUP_HOOK(init_fw_update)
```

Without the linker hook, the startup function doesn't run. If you add the linker hook then this init function runs just fine, so there's no issue with the startup hook itself.

szczys: Ah, @hasheddan here's [a different approach](https://github.com/espressif/esp-idf/tree/master/examples/build_system/cmake/plugins#ensuring-that-the-plugin-code-is-included-into-the-executable):

```
idf_component_set_property(${COMPONENT_NAME} WHOLE_ARCHIVE TRUE)
```

Can be added to the CMakeLists.txt file after the component (the user app) is registered to ensure all object files are sent to the linker. I tested and it works. Do you think this is a reasonable approach?

hasheddan: @szczys at first I was somewhat hesitant to do this because of concerns of potentially increasing code size, and the previous strategy is a bit more precise about forcing inclusion specifically. I still don't like the idea of users having to do either strategy in their application, but I don't think we have a great alternative at the moment.

I believe we should be able to mitigate the code size issues with how we conditionally build the list of source files in the Golioth SDK library (i.e. any object files in the archive we know we want to be included). We also should have any unused code that is not included in a `KEEP()` section (which are denoted in our esp-idf linker fragments) garbage collected by the linker with `--gc-sections`.

I'd probably still bias towards the more precise option, but the thing that tips the scales for me is that we fail silently if someone forgets to define the linker target function or apply the `-u` flag. I think it is probably safer to use `--whole-archive` and be (more) certain that you aren't having any future handlers stripped unexpectedly.

TL:DR: I'm good with moving forward with this approach, but let's open a tracking issue with some of this context to address 👍🏻 

(side note: this old [issue](https://github.com/zephyrproject-rtos/zephyr/issues/20663) and [PR](https://github.com/zephyrproject-rtos/zephyr/pull/28486) in Zephyr have some helpful related discussion about unconditionally including kernel code)

hasheddan (approved):
Thanks @szczys!

trond-snekvik (approved):

mniestroj (approved):

PR #183, opened 2026-04-24
Port: ESP-IDF: Package the HTTP client transport

szczys:
Move the HTTP client transport into the `port/esp_idf/transport` directory and built it as part of the Pouch component based on Kconfig values.

Port specific transport header files should now be included form the `port/<platform>include` folder.

Resolves: https://github.com/golioth/firmware-issue-tracker/issues/955


port/esp_idf/components/pouch/CMakeLists.txt
```diff
@@ -218,3 +231,6 @@ idf_component_register(SRCS
                          mbedtls
                          zcbor
                        )
+
+# Conditionally required components need to be process after initial registration
```
hasheddan:
```suggestion
# Conditionally required components need to be processed after initial registration
```

hasheddan (approved):
Thanks @szczys! One minor comment.

PR #185, opened 2026-04-24
zephyr: http_client: simplify the uplink fill step

szczys:
Reduces complexity of performing a chunk upload by separating the chunk header/footer from the pouch fill call and removing a proxy for the fill function.

However, the need for checking for 0 data length and briefly sleeping has not been resolved by this commit.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/964

### DIFF
--- a/port/zephyr/transport/http/client.c
+++ b/port/zephyr/transport/http/client.c
@@ -31,11 +31,6 @@ LOG_MODULE_REGISTER(pouch_http);
 #define HTTP_PATH_SERVER_CERT "/.g/server-cert"
 #define HTTP_PATH_DEVICE_CERT "/.g/device-cert"
 
-#define POUCH_FILL_COUNT_LIMIT 100
-#define CHUNK_DATA_LEN_FMT "%04zx"
-#define CHUNK_HEADER_LEN 6  // ascii hex len(4) + CRLF(2)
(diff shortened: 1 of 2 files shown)


port/zephyr/transport/http/client.c
```diff
@@ -268,51 +196,73 @@ static int pouch_http_payload_callback(int sock, struct http_request *req, void
...
             return -EINVAL;
         }
 
-        tx_len = write_chunk_header_footer(ctx->scratch, pouch_data_len);
+        if (0 == pouch_data_len && POUCH_MORE_DATA == res)
+        {
+            LOG_DBG("Pouch supplied 0 bytes with more data coming");
+            k_msleep(1);
+            continue;
+        }
```
szczys: We shouldn't need to check for 0 length or sleep here, but removing either causes the chunk uplink to fail because a 0-length chunk indicates the end of the post and removing the sleep results in a deadlock (`k_yield()` doesn't help).

IMO, `pouch_uplink_fill()` should never return 0 when there is more data available. This behavior is not present in the ESP-IDF port, making me wonder if there's thread priority issue at play here.

szczys: This was resolved by #189

hasheddan (approved):
Thanks @szczys!

PR #186, opened 2026-04-27
Remove unencrypted mode

trond-snekvik:
Removes the encryption type part of the CBOR header, as well as the corresponding code for handling unencrypted versions of the pouch.

Disables running the downlink test, and removes plaintext introspection test cases for the uplink test, with the expecation that we'll add a thorough test cases with sample data from the architecture docs for encryption and decryption at a later point. For now, we don't have the infrastructure to test these cases on native sim, although the encryption and decryption parts work just fine.

---

Opening in draft, as there are a couple of aspects of this change that has external dependencies:
- Do we bump the version number? This is a breaking change in the header format, but we've talked about breaking this deliberately. If we don't bump it, we're bricking existing devices.
- Are we happy with the header format, or are there other changes we want to make? We could technically flatten it, which would save us two bytes of array headers in each pouch. Now would be the time to make this change, if we want to make it.
- We need to add support for this new format in the backend and the pouch CLI util as well


tests/pouch/downlink/testcase.yaml
```diff
@@ -1,4 +1,5 @@
 common:
+  build_only: true # TODO: Disabled running this for now, as unencrypted mode is no longer supported.
```
trond-snekvik, line 2: Note this change here: I want to add sample data for uplink and downlink pouches to the spec, based on a known set of demo keys (e.g. the ones added to tests/common/src/crypto.c), so we can validate the resulting uplink and downlink formally with a set of test vectors. This test is based on introspection of plaintext pouches, so I suggest we move the test data we're creating in the script for this test to the spec source, and hardcode the resulting pouches into this test, to ensure that we only have one source of truth for the expected format.

PR #187, opened 2026-04-27
esp-idf: replace link hooks with WHOLE_ARCHIVE option

szczys:
Replace the link hook symbols used to ensure files are submitted to the linker with the WHOLE_ARCHIVE option which instructs the build system to share all object files with the linker, even if there are no globally referenced symbols in those object files.

The linker still runs garbage collection when this option is selected, so unused code will still not be added to the final binary. But the code added by iterable sections will be included in the build because our linker fragment files use KEEP().


trond-snekvik (approved):
This is the way.

mniestroj (approved):

hasheddan (approved):
Thanks @szczys!

PR #188, opened 2026-04-27
tools: create separate CI requirements and pin hashes

hasheddan:
Adds a new requirements-ci.in and uses it to generate dependencies with
pinned hashes using uv pip compile.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>


szczys (approved):

trond-snekvik (approved):

mniestroj (approved):

mniestroj:
I am just slightly worried about pinning everything.

Doing it comes at a cost, which means that we need to manually test whether we can bump (and whether we should) all the dependencies. Another thing is that we restrict which version can be installed to a single version (not a version range), regardless of other dependencies. This does not work well if every tool has a **specific** version of requirement, since there will be conflicts for sure.

hasheddan (commented):
> Doing it comes at a cost, which means that we need to manually test whether we can bump (and whether we should) all the dependencies. Another thing is that we restrict which version can be installed to a single version (not a version range), regardless of other dependencies. This does not work well if every tool has a specific version of requirement, since there will be conflicts for sure.

Chatted with @mniestroj about this synchronously, and I think this is a very valid concern. There are a couple of reasons why I am interested in pinning to specific versions:

- It ensures that our builds don't start breaking due to silent version bumps (this just happened on `main` in https://github.com/golioth/pouch/actions/runs/25057311748)
- It ensures that we aren't pulling in new, potentially compromised versions unexpectedly, exposing us to supply chain risks (ideally we would be pinning to hashes across the board)

However, we should be able to accomplish these goals without increasing the liklihood of users hitting incompatibilities. I've updated this PR to introduce a `requirements-ci.in` file, which sources the `requirements.txt` and add packages required for CI, then is used by `uv pip compile` to generate a lock file (`requirements-ci.txt`) that is then used in CI.

There are still packages being installed that are not pinned to a hash (namely all of the packages that are installed by Zephyr), which we should be able to address in a similar manner, though it is slightly more involved, so I have opted not to include in this PR.

szczys (approved):

mniestroj (approved):
I like it a lot, thanks for update! Curious how the maintenance will turn out to be in practice :)

trond-snekvik (approved):

PR #189, opened 2026-04-27
Uplink: add pouch_wait_for_queue()

szczys:
When a transport is consuming uplink blocks faster than they are being produced, `pouch_wait_for_queue()` may be used to wait for the next block using a configurable timeout.

### DIFF
--- a/src/uplink.c
+++ b/src/uplink.c
@@ -47,6 +47,8 @@ struct pouch_uplink
         pouch_buf_queue_t queue;
         /** Buffer reader for the buffer currently being processed. */
         struct pouch_bufview reader;
+        /** Semaphore to signal available blocks in the queue */
+        pouch_sem_t has_queue_sem;
     } transport;
 };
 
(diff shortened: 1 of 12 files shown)


trond-snekvik (approved):

mniestroj (approved):

hasheddan (approved):

szczys:
I updated the final commit on this PR so that we wait for the pouch to be closed before giving the semaphore. This prevents a thread waiting on the semaphore from calling `pouch_uplink_fill()` after the final block has been submitted but before the pouch is closed which affects logic in `pouch_uplink_fill()`.

src/uplink.c
```diff
@@ -85,11 +89,18 @@ static void process_blocks(pouch_work_t *work)
...
+        queue_was_updated = true;
     }
 
     if (pouch_is_closing() && !stream_is_open())
     {
         pouch_atomic_set_bit(uplink.flags, POUCH_CLOSED);
+        queue_was_updated = true;
+    }
+
+    if (true == queue_was_updated)
```
trond-snekvik: As we can't trust the semaphore to map perfectly to whether there's available data or not, this additional complexity shouldn't be added imo. We have to account for occasions where the wait function returns, but there's no buffer anyway, so the conditional setting can't really help us

trond-snekvik (approved):
LGTM, functionally, but we should just keep the signalling as simple as possible

hasheddan (approved):

PR #190, opened 2026-04-28
zephyr: coap: new pouch transport

mniestroj:
Add coap transport, that tries to map how http transport is implemented
now.

Add coap_client example that allows to utilize "pouch over coap" as a
transport method for cloud communication.

Reuse http_client pytest implementation by moving it to examples/pytest/
and pointing to it from http_client and coap_client example applications.


examples/zephyr/coap_client/Kconfig
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) 2026 Golioth, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source "$ZEPHYR_BASE/Kconfig"
+
```
szczys, line 6:
```suggestion

menu "Example: Pouch CoAP Client"

```

This looks a little nicer in `menuconfig` if wrapped in a menu. I have also opened #201 to update the http_client example in the same way.

examples/zephyr/coap_client/Kconfig
```diff
@@ -0,0 +1,65 @@
...
+    help
+        OTA component data is written to flash and mcuboot upgrade is requested.
+
+config EXAMPLE_FW_UPDATE_SHA256_VERIFY
+    bool "Download only and verify SHA256"
+    help
+        OTA component data is hashed with SHA256 instead of written
+        to flash. The resulting hash is logged to the console for verification.
+
+endchoice
```
szczys, line 67:
```suggestion
endchoice
endmenu # Example: Pouch CoAP Client
```

examples/zephyr/coap_client/prj.conf
```diff
@@ -0,0 +1,90 @@
...
+CONFIG_PSA_WANT_ALG_GCM=y
+CONFIG_PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE=y
+CONFIG_PSA_WANT_ALG_RSA_PKCS1V15_SIGN=y
+CONFIG_PSA_WANT_ALG_TLS12_PRF=y
+
+# Configure Pouch
+CONFIG_POUCH=y
+CONFIG_POUCH_TRANSPORT_COAP_CLIENT=y
+CONFIG_GOLIOTH=y
+CONFIG_GOLIOTH_SETTINGS=y
```
szczys, line 89: Just confirming that `CONFIG_GOLIOTH_OTA` is not selected by default for this example.

port/zephyr/include/pouch/transport/coap/client.h
```diff
@@ -0,0 +1,31 @@
...
+ * @return 0 Success
+ * @return non-zero Error code
+ */
+int pouch_coap_client_init(sec_tag_t sec_tag);
+
+/**
+ * Initiate the Pouch synchronization process
+ *
+ * Send uplink data to server and retrieve downlink data.
+ *
```
szczys, line 26:
```suggestion
 *
 * All CoAP state is accessed from a single thread.
 * The caller must ensure that pouch_coap_client_sync() is not
 * called concurrently.
 * 
```

port/zephyr/transport/coap/client.c
```diff
@@ -0,0 +1,673 @@
...
+    }
+
+    /* Auto-ACK confirmable responses */
+    if (coap_header_get_type(resp) == COAP_TYPE_CON)
+    {
+        int ack_ret = pouch_coap_send_ack(coap_header_get_id(resp));
+
+        if (ack_ret < 0)
+        {
+            LOG_WRN("Failed to send ACK: %d", ack_ret);
```
szczys, line 136: Do we need to return an error code here?

port/zephyr/transport/coap/client.c
```diff
@@ -0,0 +1,673 @@
...
+    }
+
+    return 0;
+}
+
+/*
+ * Fill a buffer with data from the pouch uplink, polling until
+ * the buffer is full or no more data is available.
+ */
+static int pouch_coap_fill_uplink_buffer(struct pouch_uplink *uplink,
```
szczys: The approach used in http_client was changed in #185 and removes the k_sleep in this fuction for a `pouch_wait_for_queue()` added in #189. I think we should use that approach here if possible.

szczys (commented):
Nice work @mniestroj! I left a couple of comments.

port/zephyr/transport/coap/client.c
mniestroj: The response payload has already been received and parsed successfully at this point. Failing to deliver the ACK does not invalidate the data we’re about to return to the caller. 

Propagating the ACK send failure here would just cause `pouch_coap_send_and_recv()` to retransmit the request unnecessarily, as we already hold a valid response. The CoAP server will retransmit the CON if the ACK is lost on the wire, and that retransmission will be handled (token mismatch / duplicate) by the existing recv loop.

If the socket is broken, we will fail the flow anyway.

szczys: I think we're talking about different things here. My concern was the `k_sleep(M_SEC(100))` approach to waiting for the queue.

It looks like this was fixed when you force-pushed from f054ec1 to 17f4026 so this no longer an issue.

szczys (approved):
Great work on this @mniestroj!

There is still one small question open but not a blocker.

port/zephyr/transport/coap/blockwise.c
```diff
@@ -0,0 +1,454 @@
...
+
+        /* Advance Block1 upload state */
+        if (need_block1)
+        {
+            if (send_blk_ctx.current == 0)
+            {
+                block_size_bytes =
+                    pouch_coap_negotiate_block1_size(&send_blk_ctx, &resp, block_size_bytes);
+            }
+            send_blk_ctx.current += chunk_len;
```
trond-snekvik: What does the server do if the first block is larger than its accepted block size? Presumably, it wouldn't have been able to receive the full block?

port/zephyr/transport/coap/blockwise.c
```diff
@@ -0,0 +1,454 @@
...
+
+    enum coap_block_size server_bs = block1_opt & 0x07;
+
+    if (server_bs < blk_ctx->block_size)
+    {
+        blk_ctx->block_size = server_bs;
+        return coap_block_size_to_bytes(server_bs);
+    }
+
+    return current_block_bytes;
```
trond-snekvik: When this happens, we're depending on the `current_block_bytes` actually matching `blk_ctx->block_size`, which depends on init being called with that specific argument. 
It could potentially be safer and slightly simpler to always just return `coap_block_size_to_bytes(blk_ctx->block_size)` from this function instead of passing in the `current_block_bytes`

port/zephyr/transport/coap/client.c
```diff
@@ -0,0 +1,668 @@
...
+/*
+ * Buffer size for CoAP packets.
+ * Header (~48 bytes) + block payload (up to POUCH_COAP_BLOCK_SIZE).
+ */
+#define COAP_HEADER_OVERHEAD 48
+#define COAP_BUF_SIZE (COAP_HEADER_OVERHEAD + CONFIG_POUCH_COAP_BLOCK_SIZE)
+
+static atomic_t cert_flow_flags;
+#define SERVER_CERT_DOWNLOADED_BIT BIT(0)
+#define POUCH_CERT_UPLOADED_BIT BIT(1)
```
trond-snekvik:
```suggestion
#define SERVER_CERT_DOWNLOADED_BIT (0)
#define POUCH_CERT_UPLOADED_BIT (1)
```
when used with the `atomic_set/clear_bit()` APIs

port/zephyr/transport/coap/client.c
```diff
@@ -0,0 +1,668 @@
...
+            err = -ENOMEM;
+            break;
+        }
+
+        chunk_len = sizeof(pouch_uplink_payload) - total_len;
+
+        err = pouch_coap_fill_uplink_buffer(uplink,
+                                            &pouch_uplink_payload[total_len],
+                                            &chunk_len,
+                                            &is_last);
```
trond-snekvik: Having to prefill our buffer before starting the uplink puts a pretty restrictive limit to the max size of the uplink. Since we have blockwise upload, we could call this function in a callback from the `pouch_coap_blockwise_post` while loop, and not need to scale the assembly buffer to accommodate the largest known payload. Or is there some timeout or other limitation that would prevent this? 

If we start using this in a gateway application, we won't really be able to estimate the required uplink size, but we've sort of done the ground work for being able to accommodate any size in blockwise.c, so it looks like a relatively small lift to get an infinite uplink size.

port/zephyr/transport/coap/client.c
```diff
@@ -0,0 +1,668 @@
...
+        if (err)
+        {
+            break;
+        }
+
+        total_len += chunk_len;
+    }
+
+    pouch_uplink_finish(uplink);
```
trond-snekvik, line 613: This isn't really true yet at this point, and it triggers the `POUCH_EVENT_SESSION_END` event emit that the application can subscribe to. The application could use this as a trigger to turn off its modem as a power saving mechanism - but it'd be premature, as the uplink hasn't actually gone up yet. I think this should happen together with the `downlink_finish` after the uplink has made its way out of the device, and the server has acked it

port/zephyr/transport/coap/client.c
```diff
@@ -0,0 +1,668 @@
...
+        }
+
+        state->downlink_len += len;
+    }
+
+    if (last_block && state->downlink_started)
+    {
+        pouch_downlink_finish();
+        state->downlink_started = false;
+    }
```
trond-snekvik: Since this function won't be called after `pouch_coap_blockwise_post` returns, we could just always do this at that point, instead of keeping a flag and conditionally doing it if something went wrong.

port/zephyr/transport/coap/coap_internal.h
```diff
@@ -0,0 +1,51 @@
...
+ * Functions provided by blockwise.c
+ *------------------------------------------------*/
+
+int pouch_coap_blockwise_get(const char *path, uint8_t *buf, size_t buf_size, size_t *out_len);
+
+int pouch_coap_blockwise_post(const char *path,
+                              const uint8_t *payload,
+                              size_t payload_len,
+                              block2_payload_cb_t resp_cb,
+                              void *user_data);
```
trond-snekvik, line 84: Just as a side note to the suggestion about moving the payload assembly into a callback - we could still have this variant of this function in this header, if we just implement it as a wrapper that serves a static buffer from a local callback. Would save us some boilerplate for the certificate upload function.

trond-snekvik (commented):
Sorry for the delay on this. I'm a little too reliant on the github notification workflow.

port/zephyr/transport/coap/client.c
mniestroj: The response payload has already been received and parsed successfully at this point. Failing to deliver the ACK does not invalidate the data we’re about to return to the caller.

Propagating the ACK send failure here would just cause pouch_coap_send_and_recv() to retransmit the request unnecessarily, as we already hold a valid response. The CoAP server will retransmit the CON if the ACK is lost on the wire, and that retransmission will be handled (token mismatch / duplicate) by the existing recv loop.

If the socket is broken, we will fail the flow anyway.

(I posted that comment previously to another [comment thread,](https://github.com/golioth/pouch/pull/190#discussion_r3303600865) sorry for that)

port/zephyr/transport/coap/client.c
mniestroj: Added a PR for HTTP as well: https://github.com/golioth/pouch/pull/216

port/zephyr/transport/coap/client.c
mniestroj: Good catch! Removed the early `end_session()` from `pouch_uplink_fill()` in https://github.com/golioth/pouch/pull/217. `POUCH_EVENT_SESSION_END` is now emitted only from `pouch_uplink_finish()`, which we call (together with `pouch_downlink_finish()`) after the blockwise POST completes and the server has acked the upload.

port/zephyr/transport/coap/client.c
mniestroj: Good feedback! Switched the uplink to a streaming callback (pouch_uplink_chunk_cb fed into pouch_coap_blockwise_post_streaming). Blocks are filled one at a time, so the total uplink size no longer needs to be known up front and isn't bounded by an assembly buffer.

port/zephyr/transport/coap/blockwise.c
mniestroj: Per [RFC 7959 §2.5](https://datatracker.ietf.org/doc/html/rfc7959#section-2.5), a server that accepts the block but wants a smaller subsequent size returns a Block1 option with smaller SZX. In that case server has accepted the bytes already sent and only requires subsequent blocks to use the smaller size. If server returns 4.13 it can't accept the first block at all, which we reject as a non-success code before negotiation. Note that we do not handle that case currently.

trond-snekvik (approved):

PR #191, opened 2026-04-28
requirements: pin cbor2<6.0.0 for zcbor compatibility

mniestroj:
zcbor==0.9.1 is using '>=5.4.2.post1' version, while 6.0.0 version breaks
API compatibility. Pin that to <6.0.0, so that it works.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-cbor2
+cbor2<6.0.0
 typer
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
This may be superseded by #188, but I am in favor of unblocking our CI in the meantime 👍🏻

szczys (approved):

PR #192, opened 2026-04-28
port: freertos: add Kconfig setting for work queue size

szczys:
Adds CONFIG_POUCH_FREERTOS_WORK_Q_SIZE to make the size of work queues in FreeRTOS configurable. This symbol is not honored in the Zephyr port.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/953

### DIFF
--- a/port/esp_idf/Kconfig
+++ b/port/esp_idf/Kconfig
@@ -76,6 +76,15 @@ rsource "transport/Kconfig"
 
 endif
 
+config POUCH_FREERTOS_WORK_Q_SIZE
+    int "Work queue size"
+    default 10
+    help
+        Maximum number of pouch_work_t pointers that may be added to a
+        work queue. When this limit is reached,
+        pouch_work_submit_to_queue() will return an error code until a
+        work item is processed and removed from the queue.
+
 # Libraries and SDK
 rsource "../../Kconfig"
 
--- a/port/freertos/freertos_port_layer.h
+++ b/port/freertos/freertos_port_layer.h
@@ -90,7 +90,6 @@ typedef struct freertos_sem pouch_sem_internal_t;
  * Work Queue
  *------------------------------------------------*/
 
-#define POUCH_FREERTOS_WORK_Q_DEFAULT_QUEUE_LENGTH 10
 #define POUCH_FREERTOS_WORK_FLAG_QUEUED 0
 
 typedef struct pouch_freertos_work_q pouch_work_q_internal_t;
@@ -108,7 +107,7 @@ struct pouch_freertos_work_q
 
     QueueHandle_t items;
     StaticQueue_t priv_static_queue;
-    uint8_t priv_storage[POUCH_FREERTOS_WORK_Q_DEFAULT_QUEUE_LENGTH * sizeof(pouch_work_t *)];
+    uint8_t priv_storage[CONFIG_POUCH_FREERTOS_WORK_Q_SIZE * sizeof(pouch_work_t *)];
 };
 
 struct pouch_freertos_work


mniestroj (approved):

port/zephyr/Kconfig
```diff
@@ -130,6 +130,14 @@ rsource "transport/Kconfig"
 
 endif
 
+config POUCH_WORK_Q_SIZE_LIMIT
+    int
+    default -1
+    help
+        Note: this symbol is not honored by the Zephyr RTOS port which
+        is only limited by the available memory for adding more nodes to
+        a work queue.
```
hasheddan: We already have some divergence between port configs -- I wonder if propagating this to the Zephyr port is less confusing than omitting? I generally lean towards not allowing the user to configure something that doesn't actually change behavior (i.e. show them an error if they are trying to set this config in the context of Zephyr).

szczys: Zephyr will throw an error if you try to define this symbol (see below). The problem I have is that this looks like the build is broken and there is no way for us to add more context on why this error is encountered.

> I generally lean towards not allowing the user to configure something that doesn't actually change behavior

Agreed, this Zephyr symbol cannot be configured by the user from `menuconfig`. But if they happen to have it in a prj.conf file it will not break the build. I could add a CMake warning that the symbol has no effect to further make the user aware. Or we can remove and let the Zephyr error do its thing.

So far this would be the first symbol added by Pouch that can be used by one port but not another. I'm sure there will be more situations as we implement ble support on ESP-IDF so worth some consideration now.

### The type of error I want to avoid:

```
/home/mike/canonical-compile/pouch-workspace/pouch/examples/zephyr/http_client/prj.conf:111: warning: attempt to assign the value '-1' to the undefined symbol POUCH_WORK_Q_SIZE_LIMIT
Parsing /home/mike/canonical-compile/pouch-workspace/pouch/examples/zephyr/http_client/Kconfig
Loaded configuration '/home/mike/canonical-compile/pouch-workspace/zephyr/boards/nxp/frdm_rw612/frdm_rw612_defconfig'
Merged configuration '/home/mike/canonical-compile/pouch-workspace/pouch/examples/zephyr/http_client/prj.conf'
Merged configuration '/home/mike/canonical-compile/pouch-workspace/pouch/examples/zephyr/http_client/boards/frdm_rw612.conf'
Merged configuration '/home/mike/canonical-compile/pouch-workspace/pouch/build/http_client/zephyr/.config.sysbuild'

error: Aborting due to Kconfig warnings

CMake Error at /home/mike/canonical-compile/pouch-workspace/zephyr/cmake/modules/kconfig.cmake:393 (message):
  command failed with return code: 1
Call Stack (most recent call first):
  /home/mike/canonical-compile/pouch-workspace/nrf/cmake/modules/kconfig.cmake:83 (include)
  /home/mike/canonical-compile/pouch-workspace/zephyr/cmake/modules/zephyr_default.cmake:131 (include)
  /home/mike/canonical-compile/pouch-workspace/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:66 (include)
  /home/mike/canonical-compile/pouch-workspace/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate)
  CMakeLists.txt:7 (find_package)


-- Configuring incomplete, errors occurred!
CMake Error at cmake/modules/sysbuild_extensions.cmake:530 (message):
  CMake configure failed for Zephyr project: http_client

  Location:
  /home/mike/canonical-compile/pouch-workspace/pouch/examples/zephyr/http_client
Call Stack (most recent call first):
  cmake/modules/sysbuild_images.cmake:43 (ExternalZephyrProject_Cmake)
  cmake/modules/sysbuild_default.cmake:21 (include)
  /home/mike/canonical-compile/pouch-workspace/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:75 (include)
  /home/mike/canonical-compile/pouch-workspace/zephyr/share/zephyr-package/cmake/ZephyrConfig.cmake:92 (include_boilerplate)
  /home/mike/canonical-compile/pouch-workspace/zephyr/share/sysbuild-package/cmake/SysbuildConfig.cmake:8 (include)
  template/CMakeLists.txt:10 (find_package)


-- Configuring incomplete, errors occurred!
FATAL ERROR: command exited with status 1: /usr/bin/cmake -DWEST_PYTHON=/home/mike/canonical-compile/pouch-workspace/.venv/bin/python3 -B/home/mike/canonical-compile/pouch-workspace/pouch/build -GNinja -DBOARD=frdm_rw612 -S/home/mike/canonical-compile/pouch-workspace/zephyr/share/sysbuild
```

hasheddan: Thanks for the explanation! While it is a messy error, the first line does fairly clearly communicate the issue:

```
warning: attempt to assign the value '-1' to the undefined symbol POUCH_WORK_Q_SIZE_LIMIT
```

In practice, I don't think users are going to hit this often because they would likely be copying config from an ESP-IDF project to a Zephyr project, but it certainly is possible, and your point about how we handle this generally is a good one. I'd be fine with something like `POUCH_FREERTOS_WORK_Q_SIZE` and not including on Zephyr side, then following that pattern for further divergence. What do you think?

trond-snekvik: I support this👍

mniestroj: I lean towards using `POUCH_FREERTOS_WORK_Q_SIZE` and having that only in FreeRTOS.

hasheddan (approved):
Thanks @szczys!

PR #193, opened 2026-04-29
docs: update README.md with information on ports

hasheddan:
Updates the README.md to accurately reflect the current status of platform and transport support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,40 @@
 # `pouch`
 
 > [!CAUTION]
-> `pouch` is under active development and breaking changes may be introduced at
-> any time.
+> `pouch` is under active development and breaking changes may be
+> introduced at any time.
 
-`pouch` is a non-IP protocol for communication between devices and cloud
-services, typically through one or more gateways.
+`pouch` is a transport-agnostic application layer protocol that enables
+secure and efficient transmission of data between intermittently offline
+nodes across multiple network hops. Communication over `pouch` is always
+between a **Device** and a **Broker**. A single physical device may act
+as a _gateway_ for other devices by implementing both device and broker
+support.
 
-## Adding `pouch` to a Zephyr Project
+## Platforms and Transports
+
+`pouch` is portable across platforms and transports. The following
+tables describe device and broker transport support across platforms.
+
+**Device**
+
+| Transport       | Zephyr | ESP-IDF |
+|-----------------|--------|---------|
+| BLE GATT Server | ✅     |         |
+| HTTP Client     | ✅     | ✅      |
+
+**Broker**
+
+| Transport       | Zephyr | ESP-IDF |
+|-----------------|--------|---------|
+| BLE GATT Client | ✅     |         |
+
+
+Documentation for each supported platform is provided below. While
+`pouch` may work with other platform versions, only those that are
+continuously tested and verified are listed.
+
+### Zephyr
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,13 +1,40 @@
...
-> `pouch` is under active development and breaking changes may be introduced at
-> any time.
+> `pouch` is under active development and breaking changes may be
+> introduced at any time.
 
-`pouch` is a non-IP protocol for communication between devices and cloud
-services, typically through one or more gateways.
+`pouch` is a transport-agnostic application layer protocol that enables
+secure and efficient transmission of data between intermittently offline
+entities across multiple network hops. Communication over `pouch` is
```
hasheddan: This is a bit wordy, and the term "entities" feels a little vague, but I think defining `pouch` as non-IP today is not really the right distinction given that multiple supported transports are IP based.

README.md
```diff
@@ -28,21 +56,45 @@ Python tooling. Use the following command to install it in your environment.
...
+**Supported Versions**
+
+- `v3.2.3`
+
+### ESP-IDF
+
+The `pouch` repository includes multiple [ESP-IDF
+components](https://docs.espressif.com/projects/idf-component-manager/en/latest/).
+See the [ESP-IDF examples](./examples/esp_idf) for more information on
+how to include them in an application.
```
hasheddan, line 79: I'd like to package these components so that they can be installed using the `idf.py` tooling, but it is mostly unrelated to this PR so I'm breaking it out into its own change.

szczys: I agree, we should be able to use the IDF component manager to pull in and set up Pouch as a component.

szczys (approved):

README.md
```diff
@@ -1,13 +1,40 @@
...
+| Transport       | Zephyr | ESP-IDF |
+|-----------------|--------|---------|
+| BLE GATT Server | ✅     |         |
+| HTTP Client     | ✅     | ✅      |
+
+**Broker**
+
+| Transport       | Zephyr | ESP-IDF |
+|-----------------|--------|---------|
+| BLE GATT Client | ✅     |         |
```
trond-snekvik, line 30: Love this

README.md
trond-snekvik: I think devices works better here, even though we usually just use it for embedded devices.

trond-snekvik (approved):

README.md
hasheddan: @trond-snekvik fair point -- I was trying to avoid using "devices" and then saying that one role in a pouch network is a Device (i.e. feels strange to have devices playing the Device role and devices playing the Broker role, though I don't think it is incorrect). To use more traditional network terminology, I was thinking "nodes" might be better than "entities" here. How does that sound to you?

trond-snekvik: That works for me, and should probably be defined in the architecture doc as well. E.g. "nodes are physical or virtual entities that are capable of sending and receiving pouches".

Good call out about not overloading devices, didn't really think that through.

README.md
```diff
@@ -19,6 +46,7 @@ included in any Zephyr project by adding the following to the project's
   path: modules/lib/pouch
   revision: main
   url: https://github.com/golioth/pouch.git
+  submodules: true
```
mniestroj: We don't need submodules for Zephyr.

mniestroj (approved):

PR #194, opened 2026-04-29
ESP-IDF: add unit tests to CI

szczys:
Add unit testing for the ESP-IDF port layer that runs on Linux.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/968


.github/workflows/test-esp-idf.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  unit-tests:
+    permissions:
+      contents: read
+    runs-on: ubuntu-24.04
+    container: espressif/idf:v5.4.2
```
hasheddan: Are we intentionally using `v5.4.2` here rather than `v5.4.1` in the [build action](https://github.com/golioth/pouch/blob/476eb7313904e621ae0ea337d732d40c16650895/.github/workflows/test-build-esp-idf.yml#L19)?

Also do we want to standardize on using the action vs the container? (either way lets pin to hash 👍🏻 )

szczys: I think that is a copy-pasta issue, I'm using 5.4.1 locally and it should be the same here. Will update.

I tried using the action instead of the container and it doesn't work for `idf.py --preview set-target linux`.

.github/workflows/test-esp-idf.yml
```diff
@@ -0,0 +1,49 @@
...
+              g++-multilib   \
+              libc6-dev-i386 \
+              libc6-dev:i386 \
+              libbsd-dev:i386
+
+          # Activate virtual environment
+          . $IDF_PATH/export.sh
+
+          # Python Dependencies
+          pip install -r requirements.txt
```
hasheddan:
```suggestion
          # Python Dependencies
           uv pip install \
            -r pouch/requirements-ci.txt \
            --require-hashes
```

hasheddan: I know this includes some dependencies we don't need here, but lets use the consistent install strategy 👍🏻

hasheddan (approved):
Thanks @szczys! One small suggestion below 👍🏻

hasheddan:
Ah @szczys it looks like `uv` is coming from our Zephyr container image. I’m good with using the existing pip install for now and circling back in future 👍🏻

szczys:
I think I figured out how to use UV with idf.py so I should be able to get this working with minimal hassle.

trond-snekvik (approved):

PR #195, opened 2026-04-30
CI: Run ESP-IDF using pytest

szczys:
Runs the ESP-IDF unit test using pytest. Generates a JUnit report which is retained and used to publish a summary of the run.


szczys:
@hasheddan This PR is based #194 where we have been discussing using the `requirements-ci.txt` file for python dependency. The problem I am now facing is that there are conflicting packages between test for ESP-IDF and test for Zephyr that caused bsim to fail on this run. Do you think we should create separate requirements for the two platforms or seek a different resolution?

hasheddan (commented):
> Do you think we should create separate requirements for the two platforms or seek a different resolution?

@szczys I definitely like the different requirements per platform because it aligns well with my eventual goal of pinning the hashes of the Python dependencies that come directly from Zephyr as well. We should be able to "source" the common requirements (like we do in the current `requirements-ci.in` for the `requirements.txt`), then just add on the platform-specific ones. So structure would look something like:

- `requirements.txt` (end-user facing dependencies)
- `requirements-ci.in` (common additional CI requirements -- if there are shared dependencies)
  - sources `requirements.txt`
  - (no generated `requirements-ci.txt` needed now)
- `requirements-ci-zephyr.in` (zephyr specific additional CI requirements)
  - sources `requirements-ci.in`
  - generated `requirements-ci-zephyr.txt`
- `requirements-ci-esp-idf.in` (ESP-IDF specific additional CI requirements)
  - sources `requirements-ci.in`
  - generated `requirements-ci-esp-idf.txt`
 
If the current dependencies in `requirements-ci.in` aren't needed by ESP-IDF, then we may be able to just convert it to `requirements-ci-zephyr.in`, in which case we can eliminate the shared CI requirements and just have the Zephyr and ESP-IDF variants source the `requirements.txt` directly.

szczys:
Great, I was thinking along the same lines.

hasheddan (approved):
Thanks @szczys!

PR #196, opened 2026-04-30
CI: add post-processing for ESP-IDF unit test Junit report

szczys:
- Repackage singe testsuite into individual suites by filename
- Report ignored tests as skipped
  - Unity indicates that there was a skipped test but which test was skipped is not propagated in the Junit report (only the count) and instead the test appears as passing. This commit adds a message via `TEST_IGNORE_MESSAGE` that begins with `Skipped:` which can be parsed by the post-processing script. This is brittle as it relies on the content of the message but the failure mode (if the message doesn't contain a match) is the same as not having this change.

### Before

image

### After

image

### DIFF
--- a/tests/esp_idf/port/junit_post_process.py
+++ b/tests/esp_idf/port/junit_post_process.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+import xml.etree.ElementTree as ET
+from collections import defaultdict
+from pathlib import Path
+
+
+def process_report(xml_path):
+    if not os.path.exists(xml_path):
+        print(f"Error: {xml_path} not found.")
+        return
+
+    tree = ET.parse(xml_path)
+    root = tree.getroot()
+
+    # Update the top-level <testsuites> name based on the report filename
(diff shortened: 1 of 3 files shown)


tests/esp_idf/port/junit_post_process.py
```diff
@@ -0,0 +1,109 @@
...
+
+def process_report(xml_path):
+    if not os.path.exists(xml_path):
+        print(f"Error: {xml_path} not found.")
+        return
+
+    tree = ET.parse(xml_path)
+    root = tree.getroot()  # This is the <testsuites> element
+
+    # 1. Update the top-level <testsuites> name based on the report filename
```
hasheddan: major nitpick: I wasn't totally following the numbers in the comments (some comments numbered while others not, sequence is 1,2 then 1,2,4)

hasheddan (approved):
Thanks @szczys! One minor comment below 👍🏻

tests/esp_idf/port/junit_post_process.py
szczys: Thanks for catching, I did a general cleanup on comments throughout.

PR #197, opened 2026-04-30
ESP-IDF: Add runtime provisioning to http_client example

szczys:
Resolves https://github.com/golioth/firmware-issue-tracker/issues/970


examples/esp_idf/http_client/main/CMakeLists.txt
```diff
@@ -19,36 +21,17 @@ if(NOT CMAKE_BUILD_EARLY_EXPANSION)
...
                          ${EXAMPLE_SRCS}
 
-                       EMBED_FILES
-                         ${EXAMPLE_EMBEDS}
-
                        EMBED_TXTFILES
                          ${EXAMPLE_TXT_EMBEDS}
 
+                       PRIV_REQUIRES
+                         console
```
trond-snekvik, line 34: Out of curiosity: does `PRIV_REQUIRES` make any difference when it's an application module?

examples/esp_idf/http_client/main/console.c
```diff
@@ -0,0 +1,146 @@
...
+    [CRED_TYPE_WIFI_PSK] = {.hint = USAGE_WIFI_PSK, .set_fn = cred_set_wifi_psk},
+    [CRED_TYPE_DEVICE_CRT] = {.hint = USAGE_DEVICE_CRT, .set_fn = cred_set_device_crt},
+    [CRED_TYPE_DEVICE_KEY] = {.hint = USAGE_DEVICE_KEY, .set_fn = cred_set_device_key},
+};
+
+static int set_credential(enum cred_type type, int argc, char **argv, char *log_msg)
+{
+    int err;
+
+    if (2 < argc)
```
trond-snekvik:
```suggestion
    if (2 > argc)
```

examples/esp_idf/http_client/main/console.h
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (c) 2026 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+void console_init();
```
trond-snekvik:
```suggestion
void console_init(void);
```

examples/esp_idf/http_client/README.md
```diff
@@ -2,46 +2,63 @@
...
-Optional: If you need, change the other options according to your
-requirements.
+```
+pouch> crt CRT_der_in_base64_format
+pouch> key KEY_der_in_base64_format
+```
 
-### Build, Flash, and Monitor the Project
+Device PKI credentials may be converted to base64 using your preferred
+command line tools. For example:
```
trond-snekvik, line 57: This feels a little backwards - PEM is already base64 encoded, so can we just use that instead? We're kinda forcing users to go out of their way to select DER, then make them use another command line tool to convert it to an unusual representation that is almost PEM

szczys: Agreed, this is rather convoluted right now. Here are the considerations I made when setting this up:

1. PEM files cannot be pasted to the console as their line breaks are interpreted as the end of a command (wrapping in quotes does not fix this)
2. We need DER for Pouch, but PEM is required by the HTTP client. However, support for passing DER to the client is added in ESP-IDF v6 so I think we can remove the PEM device certificates when we move to the new version
3. Our demo PKI tools provide DER and we use DER in the Zephyr version of  the examples.

I'd like to merge as-is. Or if this a big blocker for you I'd like to discuss moving to v6 and dropping PEM before merging this. This work converges with the desire to make pouch conform to the requirements of the IDF Component Manager so I don't think it'll be long before we tackle this.

examples/esp_idf/http_client/main/CMakeLists.txt
szczys: Not really, since nothing links the `REQUIRES` directive does the same thing.

szczys:
Apologies, I accidentally merged the (already approved) [temperature pr](https://github.com/golioth/pouch/pull/198) into this one. It makes not real difference beyond being annoying when reviewing this one.

trond-snekvik (approved):

examples/esp_idf/http_client/README.md
trond-snekvik: Sounds justified and well thought through :+1:

PR #198, opened 2026-05-01
examples: esp_idf: add actual temperature readings

szczys:
The ESP32 line all have internal temperature sensors. Add helper function to read that sensor and use it when streaming data.


trond-snekvik (approved):
Neat!

hasheddan (approved):

PR #203, opened 2026-05-20
Test transport SAR modules

trond-snekvik:
Adds a transport SAR test suite that tests the SAR sender and receiver by mocking endpoints and bearers.

Resolves a range of minor edge case issues, and adding enforcement of invalid behavior that would previously go unchecked.

### DIFF
--- a/src/transport/sar/packet.c
+++ b/src/transport/sar/packet.c
@@ -7,6 +7,16 @@
 #include "packet.h"
 #include <errno.h>
 
+#define TX_PKT_OFFSET_FLAGS 0
+#define TX_PKT_OFFSET_SEQ 1
+#define TX_PKT_OFFSET_FIN_CODE 1
+#define TX_PKT_OFFSET_DATA 2
+
+#define RX_PKT_OFFSET_CODE 0
+#define RX_PKT_OFFSET_SEQ 1
+#define RX_PKT_OFFSET_WINDOW 2
+
+
 #define POUCH_SAR_TX_PKT_FLAG_MASK \
     (POUCH_SAR_TX_PKT_FLAG_FIRST | POUCH_SAR_TX_PKT_FLAG_LAST | POUCH_SAR_TX_PKT_FLAG_FIN)
 
@@ -18,10 +28,16 @@ int pouch_sar_tx_pkt_decode(const void *buf, size_t buf_len, struct pouch_sar_tx
     }
 
     const uint8_t *bytes = buf;
-    pkt->flags = bytes[0] & POUCH_SAR_TX_PKT_FLAG_MASK;  // ignoring any undocumented flags
+    pkt->flags =
+        bytes[TX_PKT_OFFSET_FLAGS] & POUCH_SAR_TX_PKT_FLAG_MASK;  // ignoring any undocumented flags
     if (pkt->flags & POUCH_SAR_TX_PKT_FLAG_FIN)
     {
         if (pkt->flags != POUCH_SAR_TX_PKT_FLAG_FIN || buf_len != POUCH_SAR_TX_PKT_HEADER_LEN)
+        {
+            return -EINVAL;
+        }
+
+        if (buf_len != POUCH_SAR_TX_PKT_HEADER_LEN)
         {
             return -EINVAL;
         }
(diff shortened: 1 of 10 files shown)


trond-snekvik:
cc @soelgary - this adds a few strict checks on transport behavior that could affect the go gateway implementation

src/transport/sar/packet.c
```diff
@@ -42,7 +47,14 @@ int pouch_sar_tx_pkt_decode(const void *buf, size_t buf_len, struct pouch_sar_tx
     }
 
     pkt->seq = bytes[1];
-    pkt->data = &bytes[2];
+    if (buf_len <= POUCH_SAR_TX_PKT_HEADER_LEN)
```
hasheddan: nitpick: we shouldn't ever hit a case where `buf_len < POUCH_SAR_TX_PKT_HEADER_LEN` due to earlier check. Assuming `#define POUCH_SAR_TX_PKT_HEADER_LEN 2`, if `buf_len < POUCH_SAR_TX_PKT_HEADER_LEN`, then the access of the sequence number in the previous line would be invalid (sidenote: potentially worth using a `#define` for the sequence number index rather than just `1`). If somehow we did hit `buf_len < POUCH_SAR_TX_PKT_HEADER_LEN` here, then the following logic would be invalid anyway (though admittedly somewhat safer than than actually accessing the , so I'd lean towards using `==` to more explicitly communicate that this is an "empty" packet.
```suggestion
    if (buf_len == POUCH_SAR_TX_PKT_HEADER_LEN)
```

src/transport/sar/receiver.c
```diff
@@ -103,32 +120,88 @@ int pouch_receiver_recv(struct pouch_receiver *recv, const uint8_t *buf, size_t
...
+        {
+            POUCH_LOG_ERR("Invalid FIN (len %d)", pkt.len);
+            end(recv, false);
+            return -EINVAL;
+        }
+
+        if (recv->state == STATE_IDLE || recv->state == STATE_FAILED)
+        {
+            POUCH_LOG_ERR("Duplicate FIN");
+            return -EINVAL;
```
hasheddan, line 139: Just confirming the intention is to _not_ end in this case?

hasheddan (commented):
Awesome work @trond-snekvik! A couple of minor comments below.

src/transport/sar/packet.c
trond-snekvik: Good observations! Changed.

src/transport/sar/receiver.c
trond-snekvik: Yes. If we're already `IDLE` or `FAILED`, we've already done `end()`, and shouldn't do it again (or we're idle because we've never even started). 

There is one exception here though that I have now addressed (and added test coverage for): If the endpoint `start` function fails, we should go back to IDLE. In the reviewed version, we leave it in READY, which is not right.

src/transport/sar/packet.c
trond-snekvik, line 39: Note: There was a bug here that has been fixed in a separate commit. This is the wrong way around - the `POUCH_SAR_TX_PKT_FLAG_IDLE` should be set when we're idling as expected (and not errored out), and the new `bool success = !!(pkt.flags & POUCH_SAR_TX_PKT_FLAG_IDLE);` check in `pouch_receiver_recv` would error out incorrectly, causing the example tests to fail.

hasheddan (approved):
Nice work @trond-snekvik!

src/transport/sar/receiver.c
```diff
@@ -103,32 +126,88 @@ int pouch_receiver_recv(struct pouch_receiver *recv, const uint8_t *buf, size_t
...
+    }
+
     if (pkt.seq != ((recv->seq + 1) & POUCH_SAR_SEQ_MASK))
     {
-        POUCH_LOG_ERR("OoO RX: %x (last: %x)", pkt.seq, recv->seq);
-        end(recv, false);
-        k_work_reschedule(&recv->work, K_NO_WAIT);  // NACK
-        return -EINVAL;
+        POUCH_LOG_WRN("OoO RX: %x (last: %x)", pkt.seq, recv->seq);
+        // out of order packet - should be ignored
```
szczys, line 196: 👏 I had wondered what `OoO` meant before and now I know 😊

szczys (approved):
Great work @trond-snekvik!!

PR #204, opened 2026-05-20
Port: add delayable work abstraction

szczys:
- Add delayable work API to the port layer
- Zephyr: wrap existing delayable work with port function names
- ESP-IDF: implement delayable work functions
- Add tests for Zephyr and ESP-IDF
- Use the delayable work port in receiver.c to prepare for ESP-IDF peripheral GATT transport

resolves https://github.com/golioth/firmware-issue-tracker/issues/961


szczys:
As I originally thought, this is a tricky one to make work in both Zephyr and FreeRTOS.

I have updated the original PR to include a dedicated work queue thread for FreeRTOS delayable work. This maps to the approach that Zephyr takes by using the system work queue (which is what we're currently doing with our delayable work calls). There are some tradeoffs here:

1. The Zephyr implementation is more complex and allows for the cancel API call to remove items that have been submitted to the work queue but are not yet being processed.
    - I looked at matching this in FreeRTOS but I don't think it's worth the added complexity
    - I don't see a way to change the cancel behavior in Zephyr so it won't remove queued items
    - The compromise in this PR is to document the different behavior in the FreeRTOS port file. With our current use I don't think this difference matters.
2. We could have required that Zephyr use a custom work queue instead of the system work queue for delayable work. This would more closely map what the FreeRTOS port API is doing. However, I couldn't think of a real upside to doing this beyond parity between the ports, which we don't have anyway because of point number 1 above.
3. The ESP-IDF delayable work requires a task with a queue which means memory allocation. This is currently not used for the HTTP transport. I've added a Kconfig symbol to enabled it and guarded the implementation in the C file but not the H file. This way, memory is only allocated when actually used. The build fails only when those APIs are called without enabling the feature, and they are automatically enabled when the ble gatt transport is selected.

trond-snekvik (approved):
Great work Mike🚀  sorry for the long turn around time on this one

port/esp_idf/Kconfig
```diff
@@ -85,6 +85,39 @@ config POUCH_FREERTOS_WORK_Q_SIZE
         pouch_work_submit_to_queue() will return an error code until a
         work item is processed and removed from the queue.
 
+config POUCH_DELAYABLE_WORK
+    bool "Delayable work support"
+    default y if POUCH_TRANSPORT_BLE_GATT
+    help
+        Enable the dedicated delayable work thread (pouch_dwork).
+        This is required by transports that use the SAR receiver,
+        such as the BLE GATT transport.
```
hasheddan, line 94: It may make more sense in the future to have the transport depend on SAR, and then make that dependency select the delayable work.

hasheddan (approved):
Thanks @szczys! One non-blocking comment 👍🏻

port/esp_idf/Kconfig
szczys: Good call out. It looks like we don't currently have a symbol that selects SAR so I've made a tracking issue for adding that and mentioned the dependency change as part of that work: https://github.com/golioth/firmware-issue-tracker/issues/980

hasheddan: Perfect, thanks!

PR #205, opened 2026-05-22
ESP-IDF: Add BLE GATT Transport

szczys:
Add a BLE GATT peripheral transport to the ESP-IDF port based on NimBLE.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/976

### DIFF
--- a/port/esp_idf/components/pouch/CMakeLists.txt
+++ b/port/esp_idf/components/pouch/CMakeLists.txt
@@ -32,12 +32,12 @@ list(APPEND POUCH_PORT_SRCS
 # Pouch Core
 #########################
 
-file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/../include)
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/include)
 list(APPEND POUCH_CORE_INC_DIRS
       ${POUCH_ROOT}/include
 
       # Generated Headers
-      ${CMAKE_CURRENT_BINARY_DIR}/../include
+      ${CMAKE_CURRENT_BINARY_DIR}/include
 )
 
 list(APPEND POUCH_CORE_SRCS
@@ -81,7 +81,7 @@ if(NOT CMAKE_BUILD_EARLY_EXPANSION)
             ${CMAKE_CURRENT_BINARY_DIR}/header_decode.c
             ${CMAKE_CURRENT_BINARY_DIR}/header_encode.c
         COMMAND zcbor code -c ${POUCH_ROOT}/src/header.cddl -t pouch_header -sde
-            --include-prefix cddl/ --oc header.c --oh ../include/cddl/header.h
+            --include-prefix cddl/ --oc header.c --oh include/cddl/header.h
         BYPRODUCTS
             ${CMAKE_CURRENT_BINARY_DIR}/include/cddl/header_decode.h
             ${CMAKE_CURRENT_BINARY_DIR}/include/cddl/header_encode.h
--- a/port/esp_idf/components/pouch/CMakeLists.txt
+++ b/port/esp_idf/components/pouch/CMakeLists.txt
@@ -121,9 +121,9 @@ endif()
 if("${CONFIG_POUCH_TRANSPORT_BLE_GATT}" STREQUAL "y")
 
     list(APPEND POUCH_TRANSPORT_SRCS
-        ${POUCH_TRANSPORT_COMMON}/endpoints/info.c
-        ${POUCH_TRANSPORT_COMMON}/endpoints/uplink.c
-        ${POUCH_TRANSPORT_COMMON}/endpoints/downlink.c
+        ${POUCH_TRANSPORT_COMMON}/endpoints/device/info.c
+        ${POUCH_TRANSPORT_COMMON}/endpoints/device/uplink.c
+        ${POUCH_TRANSPORT_COMMON}/endpoints/device/downlink.c
         ${POUCH_TRANSPORT_COMMON}/sar/sender.c
         ${POUCH_TRANSPORT_COMMON}/sar/receiver.c
         ${POUCH_TRANSPORT_COMMON}/sar/packet.c
@@ -132,10 +132,16 @@ if("${CONFIG_POUCH_TRANSPORT_BLE_GATT}" STREQUAL "y")
 
     if ("${CONFIG_POUCH_ENCRYPTION_SAEAD}" STREQUAL "y")
         list(APPEND POUCH_TRANSPORT_SRCS
-            ${POUCH_TRANSPORT_COMMON}/endpoints/device_cert.c
-            ${POUCH_TRANSPORT_COMMON}/endpoints/server_cert.c
+            ${POUCH_TRANSPORT_COMMON}/endpoints/device/device_cert.c
+            ${POUCH_TRANSPORT_COMMON}/endpoints/device/server_cert.c
         )
     endif()
+
+    list(APPEND POUCH_TRANSPORT_INC_DIRS
+        ${POUCH_ROOT}/src
+        ${POUCH_TRANSPORT_COMMON}
+        ${CMAKE_CURRENT_BINARY_DIR}/include
+    )
 
     if(NOT CMAKE_BUILD_EARLY_EXPANSION)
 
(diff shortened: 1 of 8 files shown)


port/esp_idf/components/pouch/CMakeLists.txt
```diff
@@ -121,9 +121,9 @@ endif()
 if("${CONFIG_POUCH_TRANSPORT_BLE_GATT}" STREQUAL "y")
 
     list(APPEND POUCH_TRANSPORT_SRCS
-        ${POUCH_TRANSPORT_COMMON}/endpoints/info.c
-        ${POUCH_TRANSPORT_COMMON}/endpoints/uplink.c
-        ${POUCH_TRANSPORT_COMMON}/endpoints/downlink.c
+        ${POUCH_TRANSPORT_COMMON}/endpoints/device/info.c
+        ${POUCH_TRANSPORT_COMMON}/endpoints/device/uplink.c
+        ${POUCH_TRANSPORT_COMMON}/endpoints/device/downlink.c
```
trond-snekvik, line 126: Looks like I missed this in my PR, sorry about that!

port/esp_idf/transport/gatt/peripheral.c
```diff
@@ -0,0 +1,438 @@
...
+static const ble_uuid128_t pouch_downlink_uuid = POUCH_GATT_UUID_DOWNLINK_CHRC_VAL;
+static const ble_uuid128_t pouch_uplink_uuid = POUCH_GATT_UUID_UPLINK_CHRC_VAL;
+static const ble_uuid128_t pouch_info_uuid = POUCH_GATT_UUID_INFO_CHRC_VAL;
+
+#if CONFIG_POUCH_ENCRYPTION_SAEAD
+static const ble_uuid128_t pouch_server_cert_uuid = POUCH_GATT_UUID_SERVER_CERT_CHRC_VAL;
+static const ble_uuid128_t pouch_device_cert_uuid = POUCH_GATT_UUID_DEVICE_CERT_CHRC_VAL;
+#endif
+
+#define CHR_FLAGS (BLE_GATT_CHR_F_WRITE | BLE_GATT_CHR_F_NOTIFY | BLE_GATT_CHR_F_WRITE_ENC)
```
trond-snekvik, line 286: Does `BLE_GATT_CHR_F_WRITE_ENC` have an effect when we have `BLE_GATT_CHR_F_WRITE`?

port/esp_idf/transport/gatt/peripheral.c
```diff
@@ -0,0 +1,438 @@
...
+static struct pouch_characteristic info = CHAR_INIT_SENDER(&pouch_device_endpoint_info);
+
+#if CONFIG_POUCH_ENCRYPTION_SAEAD
+static struct pouch_characteristic server_cert =
+    CHAR_INIT_RECEIVER(&pouch_device_endpoint_server_cert);
+static struct pouch_characteristic device_cert =
+    CHAR_INIT_SENDER(&pouch_device_endpoint_device_cert);
+#endif
+
+static struct pouch_characteristic *all_chars[] = {
```
trond-snekvik:
```suggestion
static struct pouch_characteristic *const all_chars[] = {
```
could potentially make an `enum pouch_characteristic` with `DOWNLINK`, `UPLINK`, `INFO` and so on to just instantiate the entries in an array too, as we don't really need the separate variables.

Alternatively, if the connection handle is made into a single top level variable, we could iterate through the `pouch_gatt_svcs` characteristics array in `char_find_by_val_handle` instead.

port/esp_idf/transport/gatt/peripheral.c
```diff
@@ -0,0 +1,438 @@
...
+        .characteristics =
+            (struct ble_gatt_chr_def[]){
+                /* Downlink */
+                {
+                    .uuid = &pouch_downlink_uuid.u,
+                    .access_cb = data_write,
+                    .arg = &downlink,
+                    .val_handle = &downlink.val_handle,
+                    .flags = CHR_FLAGS,
+                },
```
trond-snekvik: Potential macro candidate

port/esp_idf/transport/gatt/peripheral.c
```diff
@@ -0,0 +1,438 @@
...
+
+        case BLE_GAP_EVENT_DISCONNECT:
+            ESP_LOGI(TAG, "Client disconnected, reason=%d", event->disconnect.reason);
+
+            for (size_t i = 0; i < POUCH_CHAR_COUNT; i++)
+            {
+                if (all_chars[i]->conn_handle != BLE_HS_CONN_HANDLE_NONE)
+                {
+                    close(all_chars[i]);
+                    all_chars[i]->conn_handle = BLE_HS_CONN_HANDLE_NONE;
```
trond-snekvik: It'd be simpler to just keep a single connection handle variable, as it's practically always going to be the same for all of them. We can't support more than one active connection anyway.  The subscription related logic below could just be a flag in this case.

port/esp_idf/transport/gatt/peripheral.c
```diff
@@ -0,0 +1,438 @@
...
+            },
+    },
+    {0}, /* Terminator */
+};
+
+/***************************************************
+ * GAP event handler
+ **************************************************/
+
+static uint16_t pouch_mtu = 23;
```
trond-snekvik: Needs a reset between connections. Also, do we have a source for this value? Should probably exist in a define or a kconfig symbol somewhere - or even in a getter function for the active connection?

port/esp_idf/transport/gatt/peripheral.c
szczys: Your suggestion sounded right to me, even after peeking at the documentation. But the two are indeed needed. BLE_GATT_CHR_F_WRITE  enables the write, and BLE_GATT_CHR_F_WRITE_ENC adds encryption to the write.

I spent quite a bit of time looking for some explanation of this but found none. However, looking at the examples, the two are always used together, or BLE_GATT_CHR_F_WRITE is used by itself. But never the other way around. For instance:

https://github.com/espressif/esp-idf/blob/c1b7f3821282cf974ee6018f9787758cc7c05ffc/examples/bluetooth/nimble/ble_phy/phy_prph/main/gatt_svr.c#L34-L35

port/esp_idf/transport/gatt/peripheral.c
szczys: Great suggestion, I changed this to use an array that is indexed by an enum value.

szczys:
I ran into some concurrency issues with what's running on the delayable work thread and what's running on the NimBLE thread. I'm setting this back to draft until I can work that out.

szczys:
I went down several rabbit holes to investigate hard faults when performing OTA with debug logging on using this transport and have several takeaways:
1. Blockbuf creation wasn't initializing the slist note which is resolved in #212 
2. ESP-IDF's debug logging is heavier than Zephyr's so I've removed hexdump debug logs from the original draft of this ble_gatt transport as the same data is generally (debug) logged by downlink.c/etc.
3. In tracking the two issue above, I may have found some issues with the uplink queue which I'll continue to investigate and address as their own issues.

This PR has incorporated @trond-snekvik's feedback, is tested as working, and ready for review. Thanks!

szczys:
> I ran into some concurrency issues with what's running on the delayable work thread and what's running on the NimBLE thread. I'm setting this back to draft until I can work that out.

Pouch uses delayable work for the ACK, so for this transport the ACK runs on the delayable work tread while all other BLE operations run on the NimBLE thread (triggered by the characteristics). I implemented an event-based system that offloaded any calls to bearer_send that didn't original from the NimBLE thread (delayable ACKs). It worked great, but I decided it was too much for what we actually need. I instead resolved stack overflows on the delayable work thread by removing debug hexdump logging from bearer_send(). That keeps this thread reliably below 2048 bytes of stack usage.

port/esp_idf/transport/gatt/peripheral.c
```diff
@@ -0,0 +1,429 @@
...
+
+/***************************************************
+ * Initialization
+ **************************************************/
+
+int pouch_gatt_init(void)
+{
+
+
+    int rc;
```
trond-snekvik, line 401:
```suggestion
{
    int rc;
```

port/esp_idf/transport/gatt/peripheral.c
```diff
@@ -0,0 +1,429 @@
...
+
+    /* Initialize characteristics */
+    for (size_t i = 0; i < POUCH_CHAR_COUNT; i++)
+    {
+        all_chars[i].subscribed = false;
```
trond-snekvik: This is done by the compiler automatically (except if we need to reinitialize)

trond-snekvik (approved):
Two minor comments, but LGTM (assuming concurrency issues can be addressed separately)!

PR #210, opened 2026-05-27
transport: sar: skip ACK resend when not connected

szczys:
send_ack() will keep retrying after receiving an error. But if there is no connection available, the ACK will continue to be rescheduled with no hope of success. Check to see if the bearer returned -ENOTCONN and skip rescheduling when found.

### DIFF
--- a/src/transport/sar/receiver.c
+++ b/src/transport/sar/receiver.c
@@ -54,6 +54,12 @@ static void send_ack(pouch_work_delayable_t *dwork)
     int err = pouch_bearer_send(p->bearer, buf, sizeof(buf));
     if (err)
     {
+        if (err == -ENOTCONN)
+        {
+            POUCH_LOG_DBG("TX skipped (%d), not connected", err);
+            return;
+        }
+
         // try again later:
         POUCH_LOG_ERR("TX failed (%d)", err);
         schedule_ack(p);


src/transport/sar/receiver.c
```diff
@@ -54,6 +54,12 @@ static void send_ack(pouch_work_delayable_t *dwork)
     int err = pouch_bearer_send(p->bearer, buf, sizeof(buf));
     if (err)
     {
+        if (err == -ENOTCONN)
+        {
+            POUCH_LOG_DBG("TX skipped (%d), not connected", err);
+            return;
+        }
```
hasheddan, line 61: @szczys what case are you seeing us encountering this scenario? I would expect that if the connection is lost we would be closing the receiver and canceling the work queue. However, if we do not close the receiver, then I suspect that we would want to not just ignore the failed ack send, but go ahead and terminate the transfer.

In short, if this situation is unrecoverable, then I would imagine that we should abort the transfer entirely rather than just ignoring a failure to ack. If the situation is recoverable, then we want to keep acking.

Or is this just to account for the case in which this ack is already in progress when the work queue is canceled?

szczys: I'll get back to you on that. This may have been cause by other issues I'm currently troubleshooting.

szczys: I am no longer able to reproduce this after fixing other concurrency issues I was having, but let me describe the behavior:

This code is running on the device and if the broker disconnects, `close()` [will be called](https://github.com/golioth/pouch/blob/fb444775ecfa8c6ad49bdf36f91b9d272df74908/port/esp_idf/transport/gatt/peripheral.c#L347) in the `BLE_GAP_EVENT_DISCONNECT` case of `pouch_gatt_gap_event`. So you're right that the transfer needs to be aborted and that is indeed how the design works.

However, I was still occassionally seeing a situation where the -ENOTCONN was returned by the NimBLE stack which led to the [schedule_ack() being called](https://github.com/golioth/pouch/blob/fb444775ecfa8c6ad49bdf36f91b9d272df74908/src/transport/sar/receiver.c#L59), starting a 1Hz loop of error messages: work_handler -> -ENOTCONN -> reschedule

One possibility is that the concurrency issues I was resolving were corrupting the memory and causing ACK to run even though a disconnect had happened.

However, the ACK is put on a delayable work queue that is running at a lower priority than some of the other threads so here's a legitimate scenarion that could happen:
1. recv->work is scheduled to send the ACK
2. Timer (or no wait) expires and adds the send_ack handler to the workq
3. Broker decides to disconnect
4. Higher priority NimBLE thread gains focus and runs the disconnect
5. Disconnect cancels the delayable work, but it's already in the workq and the FreeRTOS port doesn't remove already-queued delayable work items
6. When the workq gets focus, send_ack runs and we get -ENOTCONN from NimBLE
7. 1Hz loop begins

If this is actually what's happening, it makes a case for fixing the port so that enqueued work can be removed. However, this is a significantly complex functionality to add so I'm not advocating for it.

Actually, now that I think of it, there's also a small sliver of time in the Zephyr delayable work where the same could happen. It's after the queued work has been popped from the queue but has not yet finished processing. If the work queue is preempted, then the cancel will not stop the delayable ACK work from running once focus is back on the workq thread and we'll be in the same loop.

Thank you for coming to my Ted talk.

hasheddan: @szczys thanks for writing this up! We chatted offline and the Zephyr example of work in progress was the one that I was thinking of, but it seems as though the ESP-IDF case is potentially even more likely. Regardless, my primary concern is silently not rescheduling the ack, which effectively will halt progress, without informing the transport (bearer). I'm also somewhat concerned about `pouch_bearer_send()` failing with an error other than `-ENOTCONN`, and us finding ourselves in this same ack loop.

To address the former, I _think_ we could call `end()` here to signal to the bearer that the transfer is being aborted. To address the latter, I think it could make sense to only reschedule the ack in the event that the error is `-EAGAIN` (i.e. invert the error check logic here), and assume that if the bearer can't send and doesn't return `-EAGAIN` then we should abort the transfer.

szczys:
I'm going to think on this for awhile so I've set it back to draft. Here's where we're at:
1. We should not ignore failed ACKs
2. Tearing down the connection in these cases makes sense, but might itself be a bandaid
3. We should think about how the delayable ACKs work and see if they can be more closely tied to how endpoints control program flow.

PR #212, opened 2026-05-27
blockbuf: properly initialize buffers

szczys:
New buffers must have their slist node initialized. The blockbuf_alloc was only initializing the `bytes` member. This didn't cause an issue with Zephyr because the slist nodes do not require initialization (this is a port implementation detail). However, in FreeRTOS, using an slist node without initialization is undefined behavior.

This has been resolved by creating a buf_init() function that initializes the slist node and resets the bytes value. Note that it is safe to re-initialize an slist node in FreeRTOS as long as that node is not already linked in a list. This may be important when we implement memory slabs on that platform.

### DIFF
--- a/port/esp_idf/blockbuf.c
+++ b/port/esp_idf/blockbuf.c
@@ -14,8 +14,7 @@ struct pouch_buf *blockbuf_alloc(pouch_timeout_t timeout)
         return NULL;
     }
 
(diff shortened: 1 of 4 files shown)


hasheddan (approved):
Nice catch @szczys!

mniestroj (approved):

trond-snekvik (approved):

PR #213, opened 2026-05-28
buf: add guard to pouch_buf_queue_t

szczys:
pouch_list_*() functions are not thread safe. Guard against concurrent access by adding a lock to the pouch_buf_queue_t.

This resolves the concurrency issues between the uplink_workq thread and the pouch_work thread (when apps register uplink handlers that call pouch_uplink_entry_write), and also between the uplink_workq thread and the transport thread/context.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/981

### DIFF
--- a/src/buf.c
+++ b/src/buf.c
@@ -114,23 +114,32 @@ size_t buf_trim_end(struct pouch_buf *buf, size_t bytes)
 
 void buf_queue_init(pouch_buf_queue_t *queue)
 {
-    pouch_slist_init(queue);
+    pouch_slist_init(&queue->slist);
+    pouch_mutex_init(&queue->lock);
 }
 
 void buf_queue_submit(pouch_buf_queue_t *queue, struct pouch_buf *buf)
 {
-    pouch_slist_append(queue, &buf->node);
+    pouch_mutex_lock(&queue->lock, POUCH_FOREVER);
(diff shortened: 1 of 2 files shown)


szczys:
@mniestroj I had [a timeout on the bsim pouch.gateway.lesc test](https://github.com/golioth/pouch/actions/runs/26598465082/job/78375549576) when I first opened this. I don't think it's releated to these code changes but wanted to check with you to be sure.

src/uplink.c
```diff
@@ -106,7 +121,10 @@ static void end_session(void)
 
 void uplink_enqueue(struct pouch_buf *block)
 {
+    pouch_mutex_lock(&uplink.processing.queue_lock, POUCH_FOREVER);
     buf_queue_submit(&uplink.processing.queue, block);
+    pouch_mutex_unlock(&uplink.processing.queue_lock);
```
trond-snekvik: Perhaps this should just be part of the `pouch_buf_queue_t` instead? I added a typedef there so we wouldn't be chained to the slist. We could expand that to be a struct with an slist and a mutex, and do this in the buf_queue functions.

szczys: Ah, this is a much cleaner fix. Thanks for pointing it out. I've updated the PR to take this approach.

hasheddan (approved):

PR #216, opened 2026-06-02
http: use bit number instead of bitmask for atomic_*_bit API

mniestroj:
`atomic_{clear,set,test}_bit()` APIs take bit number (e.g. `1`), instead of
bitmask (e.g. `BIT(1)`).

This does not change anything functionally, but fixes overall atomic API
use.

### DIFF
--- a/port/esp_idf/transport/http/client.c
+++ b/port/esp_idf/transport/http/client.c
@@ -34,10 +34,10 @@ struct get_server_cert_context
     size_t pos;
 };
 
-#define IN_USE_FLAG BIT(0)
-#define SERVER_CERT_DOWNLOADED_BIT BIT(1)
-#define POUCH_CERT_UPLOADED_BIT BIT(2)
-#define DOWNLINK_IN_PROGRESS BIT(3)
+#define IN_USE_FLAG 0
+#define SERVER_CERT_DOWNLOADED_BIT 1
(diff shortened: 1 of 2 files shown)


hasheddan (commented):
@mniestroj there are a few instances like this in the `esp-idf` HTTP client port as well:

https://github.com/golioth/pouch/blob/0552629b34c0c9343847078ee56c3a53af6635b7/port/esp_idf/transport/http/client.c#L37

I believe some of those are actually setting the wrong flag (e.g. passing `BIT(3)` causing the 8th bit to toggled). Can we update / fix those as well?

mniestroj:
> @mniestroj there are a few instances like this in the `esp-idf` HTTP client port as well:
> 
> https://github.com/golioth/pouch/blob/0552629b34c0c9343847078ee56c3a53af6635b7/port/esp_idf/transport/http/client.c#L37
> 
> I believe some of those are actually setting the wrong flag (e.g. passing `BIT(3)` causing the 8th bit to toggled). Can we update / fix those as well?

All are fixed now, I believe. BIT(3) was indeed setting 8'th bit, but that was consistent between "set" and "test" APIs, so everything was working so far.

trond-snekvik (approved):

hasheddan (approved):

PR #217, opened 2026-06-02
uplink: end session when transport finishes

mniestroj:
Keep pouch_uplink_fill() focused on reporting local data availability.
Returning POUCH_NO_MORE_DATA only means the pouch byte stream has been
drained into the transport, not that the transport has sent or acknowledged
the final bytes.

Move session teardown to pouch_uplink_finish() so POUCH_EVENT_SESSION_END
is emitted only after the active transport closes the uplink session. This
avoids early session-end notifications for CoAP Block1 (in development),
HTTP chunked uploads, and Bluetooth/SAR transfers where the final bytes may
still be in flight.

### DIFF
--- a/src/uplink.c
+++ b/src/uplink.c
@@ -252,8 +252,6 @@ enum pouch_result pouch_uplink_fill(struct pouch_uplink *uplink, uint8_t *dst, s
         return POUCH_MORE_DATA;
     }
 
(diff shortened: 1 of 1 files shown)


trond-snekvik (approved):
Good fix, should be more consistent in general 👍

PR #219, opened 2026-06-02
ci: zephyr: http_client: remove cohort before generated artifacts

szczys:
artifacts were not being deleted because a deployment was using them. Wait until after the cohort has been deleted and then remove the artifacts.

### DIFF
--- a/examples/zephyr/http_client/pytest/conftest.py
+++ b/examples/zephyr/http_client/pytest/conftest.py
@@ -52,7 +52,12 @@ def test_id():
 
 
 @pytest.fixture(scope="module")
-async def cohort(project, device, test_id):
+def artifacts_to_cleanup():
+    return []
+
+
+@pytest.fixture(scope="module")
+async def cohort(project, device, test_id, artifacts_to_cleanup):
     """Create a per-device, per-test cohort so OTA deployments are isolated."""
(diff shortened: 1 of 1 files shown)


examples/zephyr/http_client/pytest/conftest.py
```diff
@@ -65,11 +70,23 @@ async def cohort(project, device, test_id):
         await device.remove_cohort()
     except Exception:
         pass
-    await project.cohorts.delete(cohort.id)
+
+    try:
+        await project.cohorts.delete(cohort.id)
+    except Exception:
+        pass
```
hasheddan, line 78: Worth logging a warning here?

trond-snekvik (approved):

mniestroj (approved):

examples/zephyr/http_client/pytest/conftest.py
hasheddan: @szczys looks like log line is now on removing the device from the cohort and there is not a log line on deleting the cohort (unless GitHub is rendering this diff incorrectly)

szczys: Apologies, it should be in the right place now.

hasheddan (approved):

PR #220, opened 2026-06-03
ESP-IDF: http_client: add automatic certificate generation to pytest

szczys:
- Migrate from accepting base64 certs as env vars to accepting paths to the der files
- Convert sync functions to async to make using pouch-pytest plugin easier
- Use pouch-pytest plug to automatically generate certificates


trond-snekvik (approved):
Neat

hasheddan (approved):
Thanks @szczys!

PR #221, opened 2026-06-03
CI: ESP-IDF: add HIL test for http_client

szczys:
Add a Hardware-in-the-Loop test for the ESP-IDF http_client example that tests stream and settings.

As logs will contain secrets, the safe-upload-artifacts action is used. We don't have access to the golioth-firmware-sdk module for ESP-IDF tests so I copied the action into this repository and updated the underlying action and pinned it.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/986
resolves: https://github.com/golioth/firmware-issue-tracker/issues/974

### DIFF
--- a/.github/workflows/test-esp-idf-hil.yml
+++ b/.github/workflows/test-esp-idf-hil.yml
@@ -1,0 +1,231 @@
+name: ESP-IDF HTTP Client HIL
+
+on:
+  workflow_dispatch:
+    inputs:
+      hil_board:
+        description: Self-hosted runner label for HIL board
+        required: true
+        type: string
+      idf_target:
+        description: ESP-IDF target (for example esp32s3)
+        required: true
+        type: string
+      api_url:
+        description: API gateway URL for Golioth backend services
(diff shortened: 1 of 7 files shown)


.github/workflows/test-esp-idf-hil.yml
```diff
@@ -0,0 +1,231 @@
...
+          submodules: recursive
+
+      - name: Build firmware and merged image
+        shell: bash
+        run: |
+          . "$IDF_PATH/export.sh"
+
+          # Install Pouch dependencies
+          pip install -r requirements-ci-esp-idf.txt --require-hashes
+
```
szczys, line 75:
```suggestion

export SDKCONFIG_DEFAULTS="sdkconfig.defaults;pytest/sdkconfig.pytest.defaults"
```

Adding the extra defaults file will make the test move faster by setting the 30s default sync down to 10s.

mniestroj (approved):

hasheddan (approved):

PR #223, opened 2026-06-04
downlink: drain decrypt work before ending crypto session

mniestroj:
Downlink blocks are decrypted asynchronously on the pouch work queue by
decrypt_blocks(). When a session ends, crypto_session_end() destroys the
downlink PSA key. If decrypt work was still in flight, the worker would use
a destroyed key handle and fail with PSA_ERROR_INVALID_HANDLE (-136).

Drain any in-flight decryption in end_session() before tearing down the
crypto session. end_session() runs on the transport/bearer thread, never on
the pouch work-queue thread, so the drain cannot self-deadlock.

The drain is implemented by flushing the work queue. Add a portable
pouch_work_queue_flush() API. On Zephyr it maps to k_work_queue_drain().
FreeRTOS port submits a sentinel work item to the queue and waits on a
semaphore.

### DIFF
--- a/port/freertos/freertos_port_layer.c
+++ b/port/freertos/freertos_port_layer.c
@@ -422,6 +422,32 @@ int pouch_work_submit_to_queue(pouch_work_q_t *queue, pouch_work_t *work)
     return -ENOMEM;
 }
 
+struct work_flush_sync
+{
+    pouch_work_t work;
+    pouch_sem_t sem;
+};
+
+static void work_flush_sentinel_handler(pouch_work_t *work)
+{
+    struct work_flush_sync *sync = CONTAINER_OF(work, struct work_flush_sync, work);
+
+    pouch_sem_give(&sync->sem);
+}
+
+void pouch_work_queue_flush(pouch_work_q_t *queue)
+{
+    struct work_flush_sync sync;
+
+    pouch_work_init(&sync.work, work_flush_sentinel_handler);
+    pouch_sem_init(&sync.sem, 0, 1);
+
+    if (0 == pouch_work_submit_to_queue(queue, &sync.work))
+    {
+        pouch_sem_take(&sync.sem, POUCH_FOREVER);
+    }
+}
+
 /*--------------------------------------------------
  * Delayable Work
  *------------------------------------------------*/
--- a/port/include/pouch/port.h
+++ b/port/include/pouch/port.h
@@ -719,6 +719,21 @@ void pouch_work_queue_start(pouch_work_q_t *queue,
  */
 int pouch_work_submit_to_queue(pouch_work_q_t *queue, pouch_work_t *work);
 
+/**
+ * @brief Flush a work queue
+ *
+ * Blocks the calling thread until every work item pending on the queue at the
+ * time of the call has finished executing. New items submitted while this
+ * function is waiting may or may not have run by the time it returns; the queue
+ * is not guaranteed to be empty on return.
+ *
+ * Must not be called from the work queue's own worker thread, or it would
+ * deadlock.
+ *
+ * @param queue The work queue to flush
+ */
+void pouch_work_queue_flush(pouch_work_q_t *queue);
+
 /*--------------------------------------------------
  * Delayable Work
  *------------------------------------------------*/
--- a/port/zephyr/zephyr_port_layer.c
+++ b/port/zephyr/zephyr_port_layer.c
@@ -269,6 +269,24 @@ int pouch_work_submit_to_queue(pouch_work_q_t *queue, pouch_work_t *work)
     return (0 <= ret) ? 0 : ret;
 }
 
+static void work_flush_sentinel_handler(pouch_work_t *work)
+{
+    ARG_UNUSED(work);
+}
+
+void pouch_work_queue_flush(pouch_work_q_t *queue)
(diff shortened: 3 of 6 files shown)


trond-snekvik (approved):
Great catch, and clean solution!

szczys (approved):
Wow, great fix!

port/freertos/freertos_port_layer.c
```diff
@@ -422,6 +422,40 @@ int pouch_work_submit_to_queue(pouch_work_q_t *queue, pouch_work_t *work)
...
+void pouch_work_queue_flush(pouch_work_q_t *queue)
+{
+    /*
+     * Submit a sentinel work item to the queue and block until it runs. The
+     * single work-queue worker clears the QUEUED flag and runs each item to
+     * completion before pulling the next, so by the time the sentinel runs every
+     * item pending when it was submitted has finished. The sync object (sentinel
+     * work + semaphore) lives on this stack and the worker does not touch the
+     * sentinel work struct after its handler returns, so this is safe.
+     */
```
hasheddan: I believe (in contrast to Zephyr solution with `k_work_queue_drain()`, which block new work items from being enqueued from non-queue thread) that calls in the freertos port to `pouch_work_submit_to_queue()` will still be enqueued successfully while we are waiting for the queue to drain. If that is the case, then I believe we could wait for the sentinel, then call `crypto_session_end()`, then have an item that was added to the queue while draining (i.e. after the sentinel) be processed and hit the invalid key error. I believe we could simulate the Zephyr behavior by having a queue level flag that allowed for blocking submissions to the queue in the freertos port.

Perhaps this is already addressed and I have missed it though?

mniestroj: `k_work_queue_drain()` indeed blocks pushing new items to the queue. However, assuming there would be another thread that would be pushing to queue, we would still be affected by race-condition on Zephyr side if we are calling with `plug=false` (which we do in current implementation). This race-condition would be in the window between the end of `k_work_queue_drain()` (`pouch_downlink_drain()`) and execution of `crypto_session_end()`.

Here in this PR I assumed that we are calling:
- pouch_uplink_finish()
- pouch_downlink_push()

both from a single thread. This is true for all transports we have today in-tree, plus with the CoAP open PR.

On the Zephyr side to allow multi-thread use, we should call `k_work_queue_drain(..., plug=true)` and then call `k_work_queue_unplug()` after we clear `SESSION_ACTIVE` atomic. The fact that Zephyr does not allow to push more items to the queue while draining is not really useful in our case, once again, assuming we use only single thread to push downlink and finish uplink.

hasheddan: @mniestroj I noticed the potential race on the Zephyr side as well, but my primary concern here is the difference in behavior of `pouch_work_queue_flush()` across the ports. Whether `plug` is set or not, on the Zephyr side, when `pouch_work_queue_flush()` returns, we know that the queue is actually empty. On the FreeRTOS side I believe we don't know because additional items may have been pushed onto the queue after the sentinel work item, so there may be items in the queue on return from `pouch_work_queue_flush()`.

mniestroj: Yes, behavior differs. IMO we have two ways forward.

A.
We can align on the behavior by implementing in Zephyr the same thing we do in ESP-IDF.

B.
Document on the API level that we do not care whether new items will be pushed to queue or not.

hasheddan: > We can align on the behavior by implementing in Zephyr the same thing we do in ESP-IDF.

I think it may be easier to implement the Zephyr behavior in FreeRTOS port by adding a queue-level flag that, when set, rejects any attempted pushes while draining. I'm not actually sure how we would mirror the FreeRTOS behavior in Zephyr (i.e. allow pushing onto the queue while draining), but perhaps you already have an idea of what that would look like?

mniestroj: I was thinking about pushing sentinel and waiting on semaphore, as we do it right now for FreeRTOS. So we would **not** use `k_work_queue_drain()` at all.

hasheddan: Notes from offline sync with @mniestroj:
- In practice, callers (transports) are not pushing things onto the queue after ending the session, so the race isn't applicable today
- In general, our goal here is to _flush_ the queue (i.e. process anything that is already on the queue, not necessarily ensure that it is fully emptied (_drain_)
- To reflect these goals (and have a simpler implementation), we'll use the sentinel work item across both Zephyr and FreeRTOS ports, and we'll rename API from `drain` to `flush`

Separately, we also discussed whether we can flip the `SESSION_ACTIVE` flag prior to flushing, but we'll treat that as a separate issue.

mniestroj: Implemented the "flush with sentinel" approach in Zephyr.

We could not map Zephyr implementation 100% with FreeRTOS (i.e. by using semaphore), but there is the same behavior on Zephyr and FreeRTOS. This is achieved by using `k_work_flush()`, which makes sure that work state (`work->flags`) is not accessed after the API call. At the end we get the same "all previously submitted work items were processed" on both OSes.

hasheddan (approved):
Nice work @mniestroj! Thanks for iterating on this one!

PR #225, opened 2026-06-04
gateway: initial pouch over coap

mniestroj:


port/zephyr/transport/coap/client.c
```diff
@@ -507,6 +515,16 @@ static int pouch_coap_upload_cert(void)
...
 }
 
+size_t pouch_coap_server_cert_get(const uint8_t **buf)
+{
+    if (buf != NULL)
+    {
+        *buf = server_cert_buf;
+    }
+
+    return atomic_get(&server_cert_len);
```
szczys, line 522: Are we intending this to return the stored length when buf is NULL?

src/gateway/uplink.c
szczys: Do we need to pass `is_last` as part of submit_wblock (or some other function). It doesn't look like it was being sent along before this PR either.

src/gateway/uplink.c
```diff
@@ -4,180 +4,192 @@
...
+    uplink->total_len += buf_size_get(uplink->wblock);
+    buf_queue_submit(&uplink->queue, uplink->wblock);
+    uplink->wblock = NULL;
+}
+
+/*
+ * Concatenate all queued blocks into a single contiguous buffer.
+ * Caller frees @p out_buf.
+ */
+static int concat_blocks(struct pouch_gateway_uplink *uplink, uint8_t **out_buf, size_t *out_len)
```
szczys: Does this relate to your commit message:
> The uplink now concatenates the full pouch into a pouch_buf and forwards it through pouch_gateway_cloud_forward_pouch() in one call instead of streaming block-by-block, dropping the bespoke block.[ch] pool in favour of pouch_buf/blockbuf.

Is there a limit to how big this might be?

src/gateway/cert.c
```diff
@@ -36,7 +29,7 @@ struct pouch_gateway_device_cert_context
 
 struct pouch_gateway_server_cert_context
 {
-    atomic_val_t id;
+    long id;
```
szczys, line 32: Is this protected?

szczys (approved):
Wow, amazing work on this @mniestroj . I really like the registration system for the transports!

src/gateway/cert.c
mniestroj: We don't have to protect that. Previously we have used atomic_t to match return value from `atomic_get()` API and now we are using `pouch_atomic_get_value()` instead, which returns `long`. So this is why we change here.

src/gateway/uplink.c
mniestroj: I have changed that to streaming, so that we do not have to accumulate whole pouches. Thanks for bringing that up!

src/gateway/uplink.c
mniestroj: I have removed that in one of the commits, it was indeed not used anymore. Good catch!

port/zephyr/transport/coap/client.c
mniestroj: I dropped the NULL check there, since there was no risk of passing NULL (we have only 1 caller).

PR #226, opened 2026-06-04
ESP-IDF: add pytest to ble_gatt example

szczys:
This adds pytest to the ble_gatt example so that testing can easily be run locally. Make sure you have a gateway running, then follow the commands in the pytest/README.md file.

This copied from the http_client with minor changes (no wifi creds, etc.). While most of the harness could be reused between the two examples, I think that work should wait until we know we will have more examples.

This prepares the example for HIL testing once we get to adding BLE in the loop (relates to #119).


examples/esp_idf/ble_gatt/pytest/sdkconfig.pytest.defaults
```diff
@@ -0,0 +1 @@
+CONFIG_EXAMPLE_HTTP_CLIENT_SYNC_PERIOD_S=10
```
hasheddan: https://github.com/golioth/pouch/blob/18f62fa147a907f5f6ba765f887f749261d95105/examples/esp_idf/ble_gatt/main/Kconfig.projbuild#L11

```suggestion
CONFIG_EXAMPLE_SYNC_PERIOD_S=10
```

Perhaps worth making this consistent across examples?

scripts/pytest-pouch/pytest_pouch/esptool_v4_shim.py
hasheddan, line 1: I presume the idea behind copying is avoid making modular before we have more data on the needs of various examples (and potential reorganization). Would something like this, which appears to be verbatim the same across both examples (and shouldn't be impacted too much by any reorganization) be a candidate for pulling out now though?

szczys: Okay, good nudge. I'll work on getting the common parts implemented as a sub-harness of the plugin so they are reusable.

hasheddan (approved):
Nice work @szczys!

PR #227, opened 2026-06-04
Transport: http_client: check status codes and cert set response

szczys:
Clean up http_client transport response handling by checking the status code for success, validating the return code from pouch_server_certificate_set(), and using the ESP-IDF http client event system to manage chunked responses.


hasheddan (approved):
Nice work @szczys!

PR #228, opened 2026-06-05
ESP-IDF: Bump ESP-IDF to v6.0.1

szczys:
Bump to newest version and remove shim for older esptool.

## Known Issues

pytest-embedded-idf calls esptool with `hard_reset` but the newer syntax is `hard-reset` resulting in this warning:
```
2026-06-05 15:05:42 Warning: Deprecated: Choice 'hard_reset' for option '--after' is deprecated. Use 'hard-reset' instead.
```

However, this does not adversely affect the CI run.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/971


src/cert.c
```diff
@@ -130,6 +133,38 @@ static int authenticate_server_cert(mbedtls_x509_crt *cert)
 
 static int extract_pubkey(mbedtls_x509_crt *cert, struct pubkey *out)
 {
+/* Add PK to PSA from mbedtls >=v4, otherwise use legacy EC functions */
+#if defined(MBEDTLS_VERSION_MAJOR) && (MBEDTLS_VERSION_MAJOR >= 4)
+    int err;
+    mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
+    psa_key_attributes_t attrs = PSA_KEY_ATTRIBUTES_INIT;
+
+    err = mbedtls_pk_get_psa_attributes(&cert->pk, PSA_KEY_USAGE_VERIFY_HASH, &attrs);
```
hasheddan, line 144: No action needed here (unless you want to add a comment), but just remarking that I find the `usage` parameter in this mbedtls API very confusing. I had to look through the [implementation](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/665b2368aabb6f381ac63251993198d192667d6d/extras/pk.c#L545) to understand exactly what a given selection implies in terms of checks and assigned attributes. It ultimately doesn't matter for us though because we are just doing this dance to export anyway.

src/cert.c
```diff
@@ -130,6 +133,38 @@ static int authenticate_server_cert(mbedtls_x509_crt *cert)
...
+    psa_reset_key_attributes(&attrs);
+
+    if (err)
+    {
+        POUCH_LOG_ERR("Extract PK: failed to import key to PSA: 0x%x", -err);
+        return -EIO;
+    }
+
+    psa_status_t status = psa_export_public_key(key_id, out->data, sizeof(out->data), &out->len);
+    (void) psa_destroy_key(key_id);
```
hasheddan: There isn't much action we can take here, and we probably don't want to fail, but perhaps worth logging an error. Depending on how PSA is configured, we could be leaking data into a fairly limited storage region (e.g. trusted storage). I'm not sure how various implementations handle all slots being full, but we could have a better chance of catching any issues if we have some error logs.

hasheddan (approved):
Nice work @szczys! One rant that requires no action, and one fairly minor suggestion.

examples/esp_idf/http_client/main/credentials.c
```diff
@@ -132,13 +125,7 @@ static int convert_device_pk_der_to_pem(char *pem_buf, size_t pem_buf_len)
...
     }
 
-    int err = mbedtls_pk_parse_key(&pk,
-                                   (unsigned char *) key_der.buffer,
-                                   key_der.size,
-                                   NULL,
-                                   0,
-                                   mbedtls_psa_get_random,
-                                   MBEDTLS_PSA_RANDOM_STATE);
+    int err = mbedtls_pk_parse_key(&pk, (unsigned char *) key_der.buffer, key_der.size, NULL, 0);
```
trond-snekvik, line 128: I see from [the migration guide](https://github.com/Mbed-TLS/mbedtls/blob/development/docs/4.0-migration-guide.md#no-random-generator-instantiation) that the RNG is removed, but for mbedTLS version < 4.0, we'll still need it, right?

hasheddan: Since we are only supporting one version of each platform at the moment, and this PR takes us to `esp-idf` `v6.0.1`, we should be safe to not support mbedTLS < 4.0 in `esp-idf` specific examples.

trond-snekvik: Fair! I thought this was in src/cert.c when I made the comment :see_no_evil: 
Rather concerning to see the amount of breaking changes in mbedTLS 4, especially with the Zephyr LTS sitting on the wrong side of that release. The reduced duplication with PSA is a welcome change though.

hasheddan: Yeah the `TF-PSA-Crypto` split is pretty severe. I do think that moving forward it will be a much better architecture though.

src/cert.c
szczys: I scratched my head on this one for a bit too long and researched how to handle filling up those key slots with failed destroy calls. But it's actually not a catastrophic issue because PSA_KEY_ATTRIBUTES_INIT selects PSA_KEY_LIFETIME_VOLATILE. We could exhaust key slots from this failure, but only for the current run since they are not persistently stored; a power cycle can clear them.

I think this is an acceptable failure mode as long as we do not return an error when psa_destroy_key() is called. If we did that, we risk a loop that fills all slots if the same thing happens again. At this point we leaked the key once, but we also got the pubkey we were after. We could retry the destroy step (I'm leaning no on that) but we should not retry the entire operation.

I'll add logging for this error in case we need it for debugging, but we don't need to retry.

src/cert.c
szczys: I added a log because this one was wicked hard for me to wrap my head around as well. Thanks for the nudge, I should have added it in the first place given the obvious reaction to reading this line of code is skepticism.

examples/esp_idf/http_client/main/credentials.c
szczys: This is generally a confusing section. I have another branch where I remove this function entirely because ESP-IDF http client can accept DER format so we no longer need to convert. That's a whole can of worms in itself so I'll save the fun surprises for a different PR.

src/cert.c
hasheddan: Nice! The comment and the log are very helpful 👍🏻

hasheddan (approved):

src/cert.c
hasheddan: Yeah even if the key lifetime wasn't volatile, there isn't really much we can do here other than attempt to notify. Log line is great 👍🏻

trond-snekvik (approved):

PR #229, opened 2026-06-08
ESP-IDF: HTTP client transport: Use DER for client credentials instead of PEM

szczys:
The original implementation used PEM formatting for the client certificate and key with an eye toward ESP-IDF v6 which added support for DER format. This PR makes that change, simplifying the credentials handling.

When working on this I discovered that DER format handling is not new, it is just (still) not documented and the configuration struct members are poorly worded.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/972


trond-snekvik (approved):
Nice. I was sort of hoping we could unify credentials.c at some point, but that would probably have to be under a platform specific directory anyway, so the cleaner the better.

hasheddan (approved):

PR #231, opened 2026-06-09
ESP-IDF: build Pouch using the idf component manager

szczys:
This implements the minimum number of changes necessary to build Pouch using the IDF component manager.
- ESP-IDF builds now originate from the root CMakeLists.txt but to limit the scope of this PR, the port cmake fail is included and the build is still handled there
- Move the existing pouch component cmake file into the port directory
- Move the existing zcbor component in the port lib directory and build as a library instead of a component
- Use idf_component.yml in the examples to build Pouch

Examples that need golioth_sdk still rely on adding an external component directory. This will be resolved in future work.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/988
resolves: https://github.com/golioth/firmware-issue-tracker/issues/989


hasheddan (approved):
Nice work @szczys! Looking forward to getting the `golioth_sdk` component moved over as well 👍🏻

trond-snekvik (approved):

PR #232, opened 2026-06-10
Add serial transport

trond-snekvik:
Adds the serial transport layer, as outlined in golioth/architecture#85. This does not include a real transport adapter, but includes three test suites that test the serial device, serial broker and a test that covers both together, each using mock serial adapters and endpoints.

There's some cleanup and some hacking in the kconfig menusystem going into this PR. The intent is to just kinda make serial build on both the gateway and the device without making our menuconfig worse. I'll make a separate PR with some more deliberate restructuring.

### DIFF
--- a/include/pouch/transport/serial/broker.h
+++ b/include/pouch/transport/serial/broker.h
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2026 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include <pouch/port.h>
+
+/**
+ * @file broker.h
+ * @brief Pouch Serial transport (broker side)
+ *
+ * The broker is the master of the serial exchange. It initiates all transfers,
+ * drives the provisioning and data exchange sequence, and owns a
+ * @ref pouch_serial_broker instance allocated by the transport adapter.
+ *
+ * Typical usage:
+ *
+ *   1. Embed @ref pouch_serial_broker inside your adapter context struct.
+ *   2. Fill in a @ref pouch_serial_broker_adapter with your ready/end callbacks
+ *      and the adapter's maximum frame size.
+ *   3. Call @ref pouch_serial_broker_init once at startup.
+ *   4. Call @ref pouch_serial_broker_start to begin an exchange with a device.
+ *   5. Feed received frames to @ref pouch_serial_broker_recv.
+ *   6. Call @ref pouch_serial_broker_notify when the device signals that it
+ *      has data available (e.g. via an interrupt line).
+ */
+
+struct pouch_serial_broker;
+struct pouch_gateway_node_info;
+
+/**
+ * Transport adapter interface for the serial broker.
+ *
+ * The adapter is responsible for all physical bus concerns (framing, length
+ * delimitation, error detection). It presents complete frames to the broker
+ * protocol layer and retrieves complete frames from it via
+ * @ref pouch_serial_broker_frame_get.
+ */
+struct pouch_serial_broker_adapter
+{
+    /**
+     * Signal to the adapter that the broker has a frame ready to send.
+     *
+     * Called whenever @ref pouch_serial_broker_frame_get would return a non-zero
+     * length. The adapter should call @ref pouch_serial_broker_frame_get at its
+     * next opportunity.
+     *
+     * @param broker Broker instance.
+     */
+    void (*ready)(const struct pouch_serial_broker *broker);
+
+    /**
+     * Called when the exchange sequence completes or fails.
+     *
+     * @param broker  Broker instance.
+     * @param success true if the full exchange completed successfully.
+     */
+    void (*end)(const struct pouch_serial_broker *broker, bool success);
+};
+
+/**
+ * Create a broker instance.
+ * The broker must be started with @ref pouch_serial_broker_start before it can
+ * process frames.
+ *
+ * @param adapter Serial adapter to use for this broker.
+ * @return Pointer to the created broker instance, or NULL on failure.
+ */
+struct pouch_serial_broker *pouch_serial_broker_create(
+    const struct pouch_serial_broker_adapter *adapter);
+
+/**
+ * Start the Pouch exchange sequence with a device.
+ *
+ * Drives the full sequence:
+ *   Info -> Server Cert (if needed) -> Device Cert (if needed) ->
+ *   Uplink drain -> Downlink delivery
+ *
+ * The adapter's @c end callback is invoked when the sequence completes or
+ * fails. Only one exchange may be in progress at a time.
+ *
+ * @param broker  Broker instance.
+ *
+ * @return 0 on success, negative error code if the exchange cannot be started.
+ */
+int pouch_serial_broker_start(struct pouch_serial_broker *broker);
+
+/**
+ * Deliver a received frame from the device to the broker transport layer.
+ *
+ * Called by the transport adapter with each complete frame received from the
+ * device. The first byte of @p frame is the encoded header; the remaining
+ * bytes are the payload.
+ *
+ * @param broker  Broker instance.
+ * @param frame   Frame bytes, starting with the 1-byte header.
+ * @param len     Total frame length in bytes (header + payload).
+ * @return 0 on success, negative error code on failure.
+ */
+int pouch_serial_broker_recv(struct pouch_serial_broker *broker, const void *frame, size_t len);
+
+/**
+ * Get a frame of data to send over the serial link.
+ *
+ * Called by the transport adapter when it needs to send data to the device.
+ * The broker fills @p buf with the next pending frame (header + payload).
+ *
+ * @param broker  Broker instance.
+ * @param buf     Buffer that should be filled with the frame to send.
+ * @param maxlen  Maximum number of bytes to write to @p buf. The broker will
+ *                not write more than this many bytes.
+ *
+ * @return Number of bytes written to @p buf, or 0 if there is no frame ready.
+ */
+size_t pouch_serial_broker_frame_get(struct pouch_serial_broker *broker,
+                                     uint8_t *buf,
+                                     size_t maxlen);
+
+/**
+ * Signal to the broker that the device has data available to send.
+ *
+ * Called by the transport adapter when the device asserts an interrupt line or
+ * otherwise indicates it has uplink data ready. The broker will issue a prompt
+ * on the uplink channel at its next opportunity.
+ *
+ * @param broker  Broker instance.
+ */
+void pouch_serial_broker_notify(struct pouch_serial_broker *broker);
+
+/**
+ * Get the adapter associated with a broker instance.
+ *
+ * @param broker  Broker instance.
+ * @return Pointer to the adapter passed to @ref pouch_serial_broker_create.
+ */
+const struct pouch_serial_broker_adapter *pouch_serial_broker_adapter_get(
+    const struct pouch_serial_broker *broker);
--- a/port/zephyr/transport/Kconfig
+++ b/port/zephyr/transport/Kconfig
@@ -5,13 +5,10 @@
 choice
   prompt "Pouch transport"
   optional
+  default POUCH_TRANSPORT_SERIAL if POUCH_TRANSPORT_SERIAL_BROKER
+  default POUCH_TRANSPORT_BLE_GATT if BT
   help
     Select the transport protocol that will carry Pouch
-
-config POUCH_TRANSPORT_NONE
-  bool "NONE"
-  help
-    No transport is selected. Useful for unit testing.
 
 config POUCH_TRANSPORT_BLE_GATT
   bool "BLE GATT"
@@ -23,6 +20,7 @@ config POUCH_TRANSPORT_BLE_GATT
 
 config POUCH_TRANSPORT_HTTP_CLIENT
   bool "HTTP"
+  depends on POUCH
   imply HTTP_CLIENT
   select EVENTS
   select NET_CONNECTION_MANAGER
@@ -32,11 +30,18 @@ config POUCH_TRANSPORT_HTTP_CLIENT
 
 config POUCH_TRANSPORT_COAP_CLIENT
   bool "CoAP"
+  depends on POUCH
   select COAP
   select NET_SOCKETS_SOCKOPT_TLS
   select NET_SOCKETS_ENABLE_DTLS
   help
     CoAP/DTLS transport for Pouch
+
+config POUCH_TRANSPORT_SERIAL
+  bool "Serial"
+  help
+    Serial transport for Pouch. Implements the core serial transport,
+    running on top of a serial transport adapter.
 
 endchoice
 
@@ -52,6 +57,7 @@ config POUCH_TRANSPORT_ACK_TIMEOUT_MS
 rsource "gatt/Kconfig"
 rsource "http/Kconfig"
 rsource "coap/Kconfig"
+rsource "serial/Kconfig"
 
 module = POUCH_TRANSPORT
 module-str = Pouch Transport
--- a/src/transport/CMakeLists.txt
+++ b/src/transport/CMakeLists.txt
@@ -5,9 +5,26 @@
 pouch_config_enabled(_pouch_encryption_saead CONFIG_POUCH_ENCRYPTION_SAEAD)
 pouch_config_enabled(_pouch_transport_sar CONFIG_POUCH_TRANSPORT_SAR)
 pouch_config_enabled(_pouch_transport_ble_gatt CONFIG_POUCH_TRANSPORT_BLE_GATT)
+pouch_config_enabled(_pouch_transport_ble_gatt_broker CONFIG_POUCH_GATEWAY_GATT)
 pouch_config_enabled(_pouch_gateway CONFIG_POUCH_GATEWAY)
+pouch_config_enabled(_pouch_transport_serial CONFIG_POUCH_TRANSPORT_SERIAL)
+pouch_config_enabled(_pouch_transport_serial_broker CONFIG_POUCH_TRANSPORT_SERIAL_BROKER)
+pouch_config_enabled(_pouch_transport_serial_device CONFIG_POUCH_TRANSPORT_SERIAL_DEVICE)
 
-if(NOT _pouch_transport_sar AND NOT _pouch_transport_ble_gatt AND NOT _pouch_gateway)
+# The device-side endpoints are only used by the transports that speak the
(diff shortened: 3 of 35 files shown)


port/zephyr/transport/Kconfig
```diff
@@ -5,23 +5,22 @@
...
+  default POUCH_TRANSPORT_SERIAL if POUCH_TRANSPORT_SERIAL_BROKER
+  default POUCH_TRANSPORT_BLE_GATT if BT
   help
     Select the transport protocol that will carry Pouch
 
-config POUCH_TRANSPORT_NONE
-  bool "NONE"
-  help
-    No transport is selected. Useful for unit testing.
-
```
szczys, line 15: I think we need this for building unit tests without the dependencies of a given transport.

port/zephyr/transport/Kconfig
```diff
@@ -30,24 +29,30 @@ config POUCH_TRANSPORT_HTTP_CLIENT
...
+  help
+    Serial transport for Pouch. Implements the core serial transport,
+    running on top of a serial transport adapter.
 
-config POUCH_TRANSPORT_ACK_TIMEOUT_MS
-    int "Pouch Transport ACK timeout (ms)"
-    default 1000
-    help
-        The timeout, in milliseconds, that a receiver will wait
-        for new packets before retransmitting an acknowledgement
```
szczys, line 50: I think symbol is [used by sar/receiver.c](https://github.com/golioth/pouch/blob/b945e2632a2b2fc95a391d3c549967e230abe0c7/src/transport/sar/receiver.c#L37)

port/zephyr/transport/Kconfig
```diff
@@ -30,24 +29,30 @@ config POUCH_TRANSPORT_HTTP_CLIENT
...
 
 rsource "gatt/Kconfig"
 rsource "http/Kconfig"
 rsource "coap/Kconfig"
+rsource "serial/Kconfig"
+
+config POUCH_TRANSPORT_SAR
+    bool
+    help
+        Selected by transports that require Segmentation and Reassembly (SAR)
```
szczys: 🎉

src/transport/serial/broker.c
```diff
@@ -0,0 +1,226 @@
...
+{
+    if (broker == NULL)
+    {
+        return -EINVAL;
+    }
+
+    POUCH_LOG_DBG("Starting broker %p", broker);
+    pouch_atomic_clear_bit(&broker->flags, FLAG_SYNC);
+    pouch_atomic_clear_bit(&broker->flags, FLAG_DOWNLINK_DONE);
+    pouch_atomic_clear_bit(&broker->flags, FLAG_UPLINK_DONE);
```
szczys, line 168: Do we need to clear the FLAG_INFO_READ flag here?

src/transport/serial/broker.c
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    pouch_serial_init(&broker->serial, serial_ready, channel_closed);
+
+    return broker;
+}
+
+void pouch_serial_broker_destroy(struct pouch_serial_broker *broker)
+{
+    free(broker);
```
szczys: Do we need any kind of teardown (ie: call `end()`) or check to ensure the broker isn't still in use?

src/transport/serial/broker.c
```diff
@@ -0,0 +1,226 @@
...
+        broker->adapter->ready(broker);
+    }
+}
+
+static void channel_ready(struct pouch_serial_broker *broker, enum pouch_serial_channel_id ch)
+{
+    pouch_serial_ch_ready(&broker->serial.channels[ch]);
+}
+
+static void next(struct pouch_serial_broker *broker)
```
szczys, line 50: This is a neat pattern 🎉

src/transport/serial/channel.c
```diff
@@ -0,0 +1,254 @@
...
+     * frame arrives (handle_data).  Just respond with an ACK so the remote
+     * sender knows it may proceed. */
+    if (ch->endpoint->send == NULL)
+    {
+        pouch_serial_ch_ready(ch);
+        return 0;
+    }
+
+    /* First prompt: open the transfer. */
+    if (!pouch_atomic_test_and_set_bit(&ch->flags, CH_FLAG_OPEN))
```
szczys, line 45: This prevents opening a transfer when one is already open.  Is it okay to silently return 0 when that happens?

src/transport/serial/channel.c
```diff
@@ -0,0 +1,254 @@
...
+        struct pouch_serial_header hdr = {
+            .channel = channel_id(ch),
+            .is_data = true,
+            .err = true,
+            .first = pouch_atomic_test_and_clear_bit(&ch->flags, CH_FLAG_FIRST_FRAGMENT),
+            .last = true,
+        };
+
+        *buf = pouch_serial_header_encode(&hdr);
+        pouch_atomic_clear_bit(&ch->flags, CH_FLAG_ERROR);
```
szczys: It's tricky to understand why the error flag is cleared here.

src/transport/serial/device.c
```diff
@@ -0,0 +1,51 @@
...
+{
+    if (ready_cb)
+    {
+        ready_cb();
+    }
+}
+
+void pouch_serial_device_init(pouch_serial_device_ready_cb_t cb)
+{
+    ready_cb = cb;
```
szczys, line 39: Do we save this because it's a singleton? I'm wondering because the transport will also store it via the init command on the next line.

port/zephyr/transport/Kconfig
trond-snekvik: Only for esp-idf, which [doesn't support the `optional` keyword](https://docs.espressif.com/projects/esp-idf-kconfig/en/latest/kconfiglib/language.html#the-choice-entry).

In Zephyr, the `optional` keyword for the `choice` means that no option has to be selected, so a token NONE option isn't necessary

port/zephyr/transport/Kconfig
trond-snekvik: Good catch! I moved it to the GATT transport, but that was not the right move

src/transport/serial/broker.c
trond-snekvik: Yep, good catch

src/transport/serial/broker.c
trond-snekvik: Potentially. I don't really have a use case for this function outside of cleaning up tests. It's just there to keep the API symmetrical. 

I think we'd need to introduce a bit of complexity and some locking mechanisms to guarantee that mid-transfer usage of this function doesn't blow something up, so I'll remove it instead. If we find a use case where we want to dynamically add and remove serial brokers in the future, I think we can reintroduce it and pay the cost.

src/transport/serial/channel.c
trond-snekvik: I think so. Acks can be sent mid-transfer when a channel is already open, so this should be considered normal usage.

src/transport/serial/channel.c
trond-snekvik: Dang, it is. The intention here is to only send the error once, but I think there's a bit of a mix up. This clear_bit here should be for `CH_FLAG_OPEN`, as we close the channel when it errors out (as indicated by the `.last = true` assignment just above this. 

The error should still be cleared (to keep the error once logic), but the right place to do it is in the error check above, which can be `pouch_atomic_test_and_clear_bit` instead

src/transport/serial/device.c
trond-snekvik: That's right, but there's a bit of extra spaghetti here to get a clean type for the `pouch_serial_device_ready_cb_t`. The `ready` function passed to `pouch_serial_init` is for our local callback function - which calls this user-provided callback. I didn't want to expose the pointer to the singleton structure in that callback for this API, so I added an extra wrapper callback. It does lead to some confusion though, so I don't feel great about it.

src/transport/CMakeLists.txt
```diff
@@ -29,38 +30,59 @@ target_include_directories(${_pouch_target} PRIVATE
...
-                --include-prefix cddl/ --output-c ${_pouch_generated_dir}/info_encode.c
-                --output-h ${_pouch_generated_dir}/include/cddl/info_encode.h
-            BYPRODUCTS
-                ${_pouch_generated_dir}/include/cddl/info_encode.h
-                ${_pouch_generated_dir}/include/cddl/info_encode_types.h
-            DEPENDS ${_pouch_transport_root}/info.cddl
-            WORKING_DIRECTORY ${_pouch_generated_dir}
+    if (_pouch_transport_serial_broker)
+        target_sources(${_pouch_target} PRIVATE
+            serial/device.c
```
mniestroj: Should that be `if (_pouch_transport_serial_device)`?

include/pouch/transport/serial/broker.h
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * The adapter's @c end callback is invoked when the sequence completes or
+ * fails. Only one exchange may be in progress at a time.
+ *
+ * @param broker  Broker instance.
+ * @param node    Gateway node info carrying provisioning state and gateway
+ *                contexts. Must remain valid until @c adapter->end is called.
+ * @return 0 on success, negative error code if the exchange cannot be started.
+ */
+int pouch_serial_broker_start(struct pouch_serial_broker *broker);
```
mniestroj: There is no `node` as parameter.

src/transport/CMakeLists.txt
trond-snekvik: Good catch, yes :+1:

src/transport/serial/channel.c
```diff
@@ -0,0 +1,254 @@
...
+    {
+        return;
+    }
+
+    if (!success)
+    {
+        pouch_atomic_set_bit(&ch->flags, CH_FLAG_ERROR);
+    }
+
+    ch->endpoint->end(&ch->bearer, success);
```
mniestroj, line 234:
```suggestion
    if (ch->endpoint->end) {
        ch->endpoint->end(&ch->bearer, success);
    }
```

We don't implement `end()` in `info.c` and `device_cert.c`.

src/transport/serial/channel.c
```diff
@@ -0,0 +1,254 @@
...
+    else if (hdr->first)
+    {
+        // Unexpected first fragment in an already-open channel
+        pouch_serial_ch_close(ch, false);
+        return -EINVAL;
+    }
+
+    if (len > 0)
+    {
+        int err = ch->endpoint->recv(&ch->bearer, payload, len);
```
mniestroj, line 104: Should we check if the `recv()` function is implemented for the endpoint that was selected (via header bytes)? What if there is a buggy frame which selects `info` or `uplink` endpoint instead?

src/transport/serial/channel.c
beriberikix: Confirming this isn't theoretical — it's fatal on real hardware.

Found while bringing up #304, which adds an rpmsg transport as an adapter on this serial core (per RFC #274 — each rpmsg message carries exactly one Pouch Serial frame). It doesn't modify the core; it's the first consumer to drive it with the real device endpoints rather than this PR's mock endpoints.

On an i.MX95 Cortex-M7, the first ACK the broker sends closes the info channel and takes the device down:

```
E: ***** USAGE FAULT *****
E:   Illegal use of the EPSR
E: r3/a4:  0x00000000 r12/ip:  0x00013880 r14/lr:  0x00010973
E: Faulting instruction address (r15/pc): 0x00000000
```

`addr2line 0x10972` resolves to `pouch_serial_ch_close` at `channel.c:235`. Since the info endpoint exists on every device, this fires on any real device.

It isn't reachable in CI because all 20 mock endpoint vtables across `tests/pouch/serial/*/src/stub_endpoints.c` define `.end`.

src/transport/serial/channel.c
```diff
@@ -0,0 +1,254 @@
...
+            .channel = channel_id(ch),
+        };
+        *buf = pouch_serial_header_encode(&hdr);
+
+        return POUCH_SERIAL_HEADER_LEN;
+    }
+
+    size_t len = maxlen - POUCH_SERIAL_HEADER_LEN;
+    enum pouch_result result = ch->endpoint->send(&ch->bearer, &buf[POUCH_SERIAL_HEADER_LEN], &len);
+    if (len == 0 && result == POUCH_MORE_DATA)
```
beriberikix, line 185: Found while bringing up the rpmsg transport in #304, an adapter on this serial core (RFC #274), running against the real device endpoints instead of this PR's mocks.

This early return is right in spirit, but `CH_FLAG_PENDING` was consumed at the top of the function (line 136, `test_and_clear_bit`) and isn't restored here — so "wait for the next call" never happens. The channel is permanently un-pending and every later `frame_get()` bails at line 137.

Worth contrasting with `sar/sender.c:94`, which has the same "no data at this time, will come back later" early return — but the SAR sender doesn't disarm itself on entry, so a transport can just re-drive it. This path both disarms itself *and* depends on `bearer_ready()` to re-arm, which the device endpoints never call: `src/transport/endpoints/device/uplink.c` implements only `start`/`send`/`end`.

It only bites when an endpoint produces data asynchronously — which is exactly the uplink, since pouch encrypts on its work queue a few hundred ms after the broker prompts the channel. Info and device-cert have their data buffered already and never reach this branch, so a session reliably completes the first three verbs and then hangs. From the Linux-side broker in #304:

```
{"level":"ERROR","msg":"session handler failed",
 "error":"read uplink: channel 4: timed out waiting for transfer"}
```

Minimal fix is to re-arm before returning:

```suggestion
    if (len == 0 && result == POUCH_MORE_DATA)
    {
        // Endpoint has no data ready to send, but expects to be called again later.
        // Re-arm so the transport polls us again; nothing else will.
        pouch_serial_ch_ready(ch);
        return 0;
    }
```

That's necessary but not sufficient by itself — it only helps if the transport re-drives the channel (the rpmsg adapter uses a short timed wait rather than blocking indefinitely on the ready signal). The alternative, and arguably the better design, is for the uplink endpoint to call `bearer->ready()` when the work queue produces a block: fully event-driven, no polling, and it fixes every transport at once. That touches `src/transport/endpoints/`, which this PR doesn't own, so I'd defer to you on the direction.

Possibly related: your note on the `clear_bit` at line 165 being "a bit of a mix up" — same function, same class of flag-lifecycle issue, so there may be a common cleanup here rather than two isolated patches.

beriberikix:
**Context:** RFC #274 proposes running Pouch over rpmsg on heterogeneous MPU+MCU SoCs, where an MCU core runs the Pouch device stack and syncs through a gateway on the Linux side. #304 implements that as an **adapter on this serial core** rather than a new transport — one Pouch Serial frame per rpmsg message, with no changes to the code in this PR.

This PR notes it "does not include a real transport adapter… each using mock serial adapters and endpoints." #304 is that adapter, and bringing it up on hardware (NXP FRDM-IMX95: Zephyr on the Cortex-M7, Linux plus a Go broker on the Cortex-A55s) exercised the core against the **real** endpoints for the first time. That turned up one defect not yet raised — commented inline at `channel.c:185` — and independently confirmed the one @mniestroj already flagged at `channel.c:234`.

Both are invisible to CI for the same structural reason, which seems worth fixing alongside them. Across all three suites the mocks in `tests/pouch/serial/*/src/stub_endpoints.c` differ from the real endpoints in exactly the two ways that matter:

1. **All 20 mock vtables define `.end`.** The real `pouch_device_endpoint_info` and `device_cert` don't, so the unguarded call at `channel.c:234` is unreachable in test but fatal on hardware — the first ACK the broker sends closes the info channel and faults the device.
2. **All three mocks share one synchronous send** — `memcpy` from a pre-filled buffer, then `return (s->tx_offset >= s->tx_len) ? POUCH_NO_MORE_DATA : POUCH_MORE_DATA;`. It never returns `POUCH_MORE_DATA` with `*dst_len == 0`. The real uplink endpoint hits that combination on every session, because pouch encrypts on its work queue a few hundred ms after the broker prompts the channel — and the branch at `channel.c:185` then deadlocks the transfer.

Two mock variants would close the gap without needing hardware: one with `.end = NULL`, and one that returns `POUCH_MORE_DATA` with zero bytes for the first N calls before producing data.

With both fixes applied, a full session completes end-to-end against `gw.golioth.io` — info, server certificate, device certificate, uplink, downlink — and the MCU registers in the Golioth console as its own device alongside the gateway. Device-side trace of the uplink path that previously hung:

```
[00:00:05.760] <dbg> pouch_serial_ch: handle_ack: ch4: sender start -> 0
[00:00:05.760] <dbg> pouch_serial_ch: ch4: no data yet, re-arming
[00:00:05.760] <dbg> saead_session: session_encrypt_block: Session key: …
[00:00:08.098] <dbg> downlink: pouch_downlink_start: Pouch downlink start
```

Ten consecutive sessions, zero errors.

src/transport/serial/channel.c
trond-snekvik: Did this, and added test coverage for it with a new `test_sender_without_end_callback` test case in test_device.c

src/transport/serial/channel.c
```diff
@@ -0,0 +1,269 @@
...
+    {
+        return 0;
+    }
+
+    // If this is a receiver channel
+    if (ch->endpoint->send == NULL)
+    {
+        struct pouch_serial_header hdr = {
+            .is_data = false,
+            .err = pouch_atomic_test_bit(&ch->flags, CH_FLAG_ERROR),
```
mniestroj:
```suggestion
            .err = pouch_atomic_test_and_clear_bit(&ch->flags, CH_FLAG_ERROR),
```

mniestroj: After a single receiver-side failure (e.g. `endpoint->recv()` returning an error, as in `test_receiver_recv_error`), that channel is permanently poisoned on this `pouch_serial` instance. The broker's next `start()` prompts INFO, immediately NACKs it, and the exchange dies before it begins. Should we clear the bit here?

src/transport/serial/channel.c
```diff
@@ -0,0 +1,269 @@
...
+            .is_data = false,
+            .err = false,
+            .channel = channel_id(ch),
+        };
+        *buf = pouch_serial_header_encode(&hdr);
+
+        return POUCH_SERIAL_HEADER_LEN;
+    }
+
+    size_t len = maxlen - POUCH_SERIAL_HEADER_LEN;
```
mniestroj, line 197: We need to make sure that `maxlen` is >= `POUCH_SERIAL_HEADER_LEN` here, so that `len` is valid.

mniestroj (commented):
2 things from my side, otherwise looks good.

src/transport/serial/channel.c
trond-snekvik: Good catch. I encountered this in the SPI branch, and came to the same conclusion, but didn't move the fix over here. Will do.

mniestroj (approved):
Awesome work on the design, implementation and refinement!

PR #233, opened 2026-06-10
examples: make firmware update OTA component configurable

hasheddan:
Moves away from using `main` as the hard-coded firmware update OTA component name and makes the value configurable with more specific defaults. Also updates CI to set to a common `ci_ota_fw` package name. Cohorts are used in CI to deployment package versions for each test.


hasheddan (commented):
Artifacts uploaded under new package successfully:

Image

trond-snekvik (approved):

mniestroj (approved):

PR #236, opened 2026-06-11
CI: update actions and ensure all are pinned to hashes

szczys:
We were getting deprecation warnings for older version of actions/checkout. I resolved them by updating to the latest release and used this as a chance to audit all of our actions, updating to latest and pinning to a hash.


hasheddan (approved):
Thanks @szczys!

PR #237, opened 2026-06-11
Combine CMake Build System

szczys:
Centralize the build for both ESP-IDF and Zephyr to begin at the root CMakeLists.txt file. Include port-specific CMakeLists.txt files for port layer and transports. All common code is included in the using standard cmake (target_sources, etc) with add_subdirectory and Kconfig guards.

Add the idf_component.yml necessary to use this as a (publishable) ESP-IDF component.

Because the gateway code is not yet cross-platform, those cmake files were left untouched.

## Manual testing

- [X] esp_idf/ble_gatt
- [X] esp_idf/http_client
- [X] zephyr/ble_gatt
- [x] zephyr/coap_client
- [X] zephyr/http_client
- [X] zephyr/gateway

resolves: https://github.com/golioth/firmware-issue-tracker/issues/992
resolves: https://github.com/golioth/firmware-issue-tracker/issues/973


cmake/pouch_target.cmake
```diff
@@ -0,0 +1,40 @@
...
+
+    set(${out_var} FALSE PARENT_SCOPE)
+endfunction()
+
+function(pouch_validate_config)
+    pouch_config_enabled(_pouch_encryption_saead CONFIG_POUCH_ENCRYPTION_SAEAD)
+    if(NOT _pouch_encryption_saead)
+        return()
+    endif()
+
```
mniestroj, line 21: In Zephyr we could probably do:
```
if (NOT CONFIG_POUCH_ENCRYPTION_SAEAD)
    return()
endif()
```
Isn't that the case for ESP-IDF? I just wonder what value does `pouch_config_enabled()` bring.

szczys: The problem I found is that ESP-IDF sets unselected symbols to `""` (empty string):
```
build/config/sdkconfig.cmake
1486:set(CONFIG_POUCH_DELAYABLE_WORK "")
```

But in Zephyr the symbol is just not present. I added a helper function that differentiates and sets the true/false to deal with the issue.

mniestroj (approved):

trond-snekvik (approved):
Very clean!

Sorry for the delay on this.

PR #240, opened 2026-06-16
golioth_sdk: isolate files and unify the cmake implementation

szczys:
Update the golioth_sdk to build using add_subdirectory(). This makes it easier to replace with SDKs for other purposes.

Header files for golioth_sdk are moved out of the root include directory and into golioth_sdk/include. The pouch/golioth/*.h prefix for these headers was changed to golioth/*.h

Due to different linker handling between platforms, linker scripts (.lf for ESP-IDF, .ld for Zephyr) are unconditionally added to the build. These files in their current form do not occupy system resources unless they are actually used in the code. This is necessary because ESP-IDF needs to have access to all linker fragments during early expansion. The WHOLE_ARCHIVE directive is used when golioth_sdk is selected to ensure files with no global references are submitted to the linker.

Zephyr has a quirk that prevents us from using target_include_directories() in a module. While those directories are declared PUBLIC, this links them to the module (pouch), not to the application that uses the module. A conditional check and call to zephyr_include_directories() is used as a workaround.

This PR replaces #234 as we decided not to build golioth_sdk as a component/module until it is moved to its own repository.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/967
resolves: https://github.com/golioth/firmware-issue-tracker/issues/990
resolves: https://github.com/golioth/firmware-issue-tracker/issues/991
closes: https://github.com/golioth/pouch/pull/234


port/esp_idf/CMakeLists.txt
```diff
@@ -205,17 +221,22 @@ idf_component_register(SRCS
...
+
+                       ${WHOLE_ARCHIVE_IF_SELECTED})
 
 # Conditionally required components need to be processed after initial registration
 idf_component_optional_requires(PRIVATE ${POUCH_PRIV_REQUIRES})
 
 if(NOT CMAKE_BUILD_EARLY_EXPANSION)
     add_subdirectory(${POUCH_PORT}/esp_idf/lib/zcbor zcbor)
+
+    set(POUCH_ACTIVE_TARGET ${COMPONENT_LIB})
```
szczys, line 240: This symbol will be widely used starting with #237.

port/zephyr/CMakeLists.txt
```diff
@@ -11,11 +11,19 @@ if(CONFIG_POUCH)
     zephyr_include_directories(../../include) # Pouch includes
 
     zephyr_library_named(pouch)
+    set(POUCH_ACTIVE_TARGET ${ZEPHYR_CURRENT_LIBRARY})
```
szczys, line 14: This symbol will be widely used starting with #237

golioth_sdk/CMakeLists.txt
```diff
@@ -2,35 +2,39 @@
...
+# most notably ESP-IDF which needs access to all linker files during early expansion.
 
-zephyr_linker_sources(SECTIONS dispatch.ld)
-
-# Settings
+if(Zephyr_DIR)
+    # Workaround Zephyr not linking target_include_directories at the app level
+    zephyr_include_directories(${CMAKE_CURRENT_LIST_DIR}/include)
+else()
+    target_include_directories(${POUCH_ACTIVE_TARGET} PUBLIC ${CMAKE_CURRENT_LIST_DIR}/include)
```
szczys, line 20: `POUCH_ACTIVE_TARGET` is set by the port CMake file.

trond-snekvik (approved):
LGTM! 🚀

hasheddan (approved):
Love seeing the `golioth` headers no longer prefixed with `pouch/` 🙌🏻

PR #245, opened 2026-06-24
ESP-IDF: use static queue for blockbuf

szczys:
Move from using dynamic allocation in blockbuf.c to using static allocation based on FreeRTOS queues. This maps well to the memory slabs used in the Zephyr implementation.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/960

### DIFF
--- a/port/esp_idf/blockbuf.c
+++ b/port/esp_idf/blockbuf.c
@@ -4,12 +4,46 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 #include <pouch/blockbuf.h>
+#include <pouch/port.h>
 #include "../../src/block.h"
+
+#include "freertos/FreeRTOS.h"
+#include "freertos/queue.h"
+
+#define BLOCKBUF_ALIGN 4
+#define BLOCKBUF_BLOCK_SIZE \
+    ROUND_UP((POUCH_BUF_OVERHEAD + MAX_PLAINTEXT_BLOCK_SIZE), BLOCKBUF_ALIGN)
+
+static struct pouch_buf *_blockbuf_pool[CONFIG_POUCH_BLOCK_COUNT];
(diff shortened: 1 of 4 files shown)


port/freertos/freertos_port_layer.h
```diff
@@ -86,6 +86,12 @@ typedef struct freertos_sem pouch_sem_internal_t;
     }                                                               \
     POUCH_APPLICATION_STARTUP_HOOK(static_semaphore_init_##name)
 
+/*------------------------------------------------
+ * Time
+ *------------------------------------------------*/
+
+int32_t pouch_timeout_to_freertos_ticks(int32_t pouch_timeout);
```
szczys, line 93: There is some code smell here because we are not using `pouch_timeout_t` as the type. To do so we'd need to move other code around to avoid circular dependencies. I thought this a better solution than copying the same function into blockbuf.c. The only real risk here is if the underlying type of pouch_timeout_t changes in the future.

port/esp_idf/blockbuf.c
```diff
@@ -20,5 +54,12 @@ struct pouch_buf *blockbuf_alloc(pouch_timeout_t timeout)
...
 {
-    free(buf);
+    if (buf == NULL)
+    {
+        return;
+    }
+
+    BaseType_t err = xQueueSend(_blockbuf_queue_handle, &buf, 0);
+    /* This can only return errQUEUE_FULL which should never happen */
+    configASSERT(pdPASS == err);
```
szczys, line 64: I hemmed and hawed on this. Asserting here is not fantastic, especially if assertions are turned off in the build, we won't catch this at all. I initially had a log here, but it's the only logging in the whole file which seems like overkill to register a tag just for this.

This is unlikely to happen. The only way I can think of is if someone frees a buffer that has already been freed. We wouldn't be able to put it into the queue (full) so we should fail hard on that anyway.

trond-snekvik (approved):

mniestroj (approved):

hasheddan (approved):

PR #251, opened 2026-06-29
CI: pin workflows and contianers to sha and assign write permissions

szczys:
- Upgrade all actions to latest release and pin by sha
- Update golioth-coverity-base to latest (it was broken before) and pin to digest sha
- Add write permissions to workflows that require them

### DIFF
--- a/.github/workflows/ci-trigger.yml
+++ b/.github/workflows/ci-trigger.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions:
+  contents: read
+  actions: write
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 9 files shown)


.github/workflows/ci-trigger.yml
```diff
@@ -22,18 +26,30 @@ jobs:
     secrets: inherit
   trigger-build-gateway:
     name: Run Gateway Build
+    permissions:
+      contents: read
+      actions: write
```
trond-snekvik, line 31: I was expecting the top level permissions to become the default for each job 🤔

trond-snekvik (approved):

hasheddan (approved):

.github/workflows/ci-trigger.yml
szczys: Yes, I expected this to be the case as well. The issue is with calling other workflows and documented in [a note here](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#supported-keywords-for-jobs-that-call-a-reusable-workflow):

> If jobs.<job_id>.permissions is not specified in the calling job, the called workflow will have the default permissions for the GITHUB_TOKEN. For more information, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions).

So here's what's going on in this file:
1. The per-job permissions here are necessary because the top level permissions don't apply to called workflows like `uses: ./.github/workflows/test-build-gateway.yml`
2. Top level permissions in this file apply to `trigger-test-summaries` job which uses actions to publish to the summary of the PR but doesn't call other workflows
3. We could have used job-specific permissions for `trigger-test-summaries` instead of the top-level if we wanted to.

PR #252, opened 2026-06-30
ESP-IDF: HTTP client transport: fix duplicate pouch_uplink_finish() and response callback event handling

szczys:
- Remove a duplicate call to pouch_uplink_finish() and add a null check to that function.
- Rework the response callback to use a single path for pouch_downlink_finish() and handle DISCONNECT and ERROR events

When we cleaned up the config related to chunked upload in https://github.com/golioth/pouch/pull/249 it unblocked the proper flow of the esp http client and exposed a double call to pouch_uplink_finish(). Because the first call reset the uplink struct to NULL, the second call crashed when trying to lock the mutex to free buffers.

While working on the above issue, I noted a TODO related to pouch_downlink_finish() and took the opportunity to fix how the GET event callbacks are handled.

### DIFF
--- a/port/esp_idf/transport/http/client.c
+++ b/port/esp_idf/transport/http/client.c
@@ -234,10 +234,7 @@ esp_err_t pouch_uplink_response_callback(esp_http_client_event_t *evt)
     {
         if (false == pouch_atomic_test_and_set_bit(&sync->flags, DOWNLINK_IN_PROGRESS))
         {
-            pouch_uplink_finish(sync->uplink);
-            ESP_LOGD(TAG, "Uplink complete");
-
-            sync->uplink = NULL;
+            /* This is the first downlink data */
             pouch_downlink_start();
         }
 
@@ -412,10 +409,8 @@ static int send_pouch_uplink(struct sync_context *sync)
     err = 0;
 
 finish_and_return:
-    if (NULL != sync->uplink)
-    {
-        pouch_uplink_finish(sync->uplink);
-    }
+    pouch_uplink_finish(sync->uplink);
+    sync->uplink = NULL;
 
     if (true == pouch_atomic_test_and_clear_bit(&sync->flags, DOWNLINK_IN_PROGRESS))
     {
--- a/port/esp_idf/transport/http/client.c
+++ b/port/esp_idf/transport/http/client.c
@@ -229,49 +229,46 @@ static int upload_pouch_cert(struct sync_context *sync)
 esp_err_t pouch_uplink_response_callback(esp_http_client_event_t *evt)
 {
     struct sync_context *sync = (struct sync_context *) evt->user_data;
+    int err = 0;
 
-    if (HTTP_EVENT_ON_DATA == evt->event_id)
+    switch (evt->event_id)
     {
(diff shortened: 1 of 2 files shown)


port/esp_idf/transport/http/client.c
```diff
@@ -229,49 +229,46 @@ static int upload_pouch_cert(struct sync_context *sync)
...
+            taskYIELD();
+            return err;
 
-        /* Data is chunked, feed the watchdog timer between each payload */
-        taskYIELD();
-    }
+        case HTTP_EVENT_DISCONNECTED:
+        case HTTP_EVENT_ERROR:
+            err = ESP_FAIL;
+            /* falls through */
```
szczys, line 259: This error condition doesn't do anything, but we do want to finish the downlink in these cases so the fall-through is the important part.

src/uplink.c
```diff
@@ -269,6 +269,11 @@ int pouch_uplink_error(struct pouch_uplink *uplink)
 
 void pouch_uplink_finish(struct pouch_uplink *uplink)
 {
+    if (uplink == NULL)
```
hasheddan:
```suggestion
    if (NULL == uplink)
```

hasheddan: Separate from the suggestion above, adding a NULL check here at all could be viewed as either overly defensive or a candidate for an assert. I think the latter could be reasonable if we viewed this as always  being a case of programmer error, but I think it is reasonable to say this is a valid path in a finish / cleanup scenario. I would like to be consistent with our convention here though, so it is worth evaluating (outside the context of this PR) if we are handling other similar scenarios with the same pattern.

port/esp_idf/transport/http/client.c
```diff
@@ -415,11 +409,7 @@ static int send_pouch_uplink(struct sync_context *sync)
     if (NULL != sync->uplink)
     {
         pouch_uplink_finish(sync->uplink);
-    }
-
-    if (true == pouch_atomic_test_and_clear_bit(&sync->flags, DOWNLINK_IN_PROGRESS))
-    {
-        /* TODO: pouch_downlink_finish(); */
+        sync->uplink = NULL;
     }
```
hasheddan: If we are taking the stance that finish functions can gracefully handle NULL, we shouldn't need to perform defensive checks at the caller (i.e. let's pick one or the other). As a side note, I don't think we can actually hit the case where `sync->uplink` is NULL after your changes in this PR.

szczys: I think you're right that we can't hit this.

I think we should make the pouch_finish_uplink() handle a NULL as the uplink pointer. Whether that means asserting or returning is open for debate. This PR makes it return silently (logging is not enabled in uplink.c) which some would say hides the problem. However, it's not something you can solve at compile time so I'd prefer if runtime is able to recover instead of crashing. We also envision customers writing their own transports which will call this function so hardening it seems appropriate.

pouch_downlink_finish() doesn't have a param so the issue is unique to uplink.

src/uplink.c
hasheddan: @szczys looks like the suggestion here is still unaddressed

szczys: @hasheddan apologies, it is updated now.

hasheddan (approved):

PR #268, opened 2026-07-09
examples: zephyr: block brokering until gateway ready

hasheddan:
Updates gateway transport readiness check to block until the transport is
ready. If we attempt to broker without having fetched a server certificate,
we will fail in perpetuity as the first operation we perform is updating
the server certificate if the SNR does not match.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

### DIFF
--- a/examples/zephyr/gateway/src/main.c
+++ b/examples/zephyr/gateway/src/main.c
@@ -407,10 +407,15 @@ int main(void)
     {
         wait_for_network();
 
-        err = pouch_gateway_cloud_ensure_ready();
-        if (err)
+        while (true)
         {
+            err = pouch_gateway_cloud_ensure_ready();
+            if (!err)
+            {
+                break;
+            }
             LOG_WRN("Initial cloud transport setup failed: %d, will retry", err);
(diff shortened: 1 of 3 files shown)


szczys (approved):

examples/zephyr/gateway/src/main.c
```diff
@@ -449,6 +454,12 @@ int main(void)
             if (err)
             {
                 LOG_WRN("CoAP sync failed: %d", err);
+                continue;
+            }
+            err = pouch_gateway_cloud_ensure_ready();
+            if (err)
+            {
+                LOG_WRN("Initial cloud transport setup failed: %d, will retry", err);
```
mniestroj: Do we need this as well? Looks like the change after `wait_for_network();` is all we need.

hasheddan: @mniestroj thanks for catching! Meant to remove this after settling on blocking until transport ready -- removed!

mniestroj (approved):

PR #270, opened 2026-07-10
{coap,http}_client: support NCS and nrf9160dk

mniestroj:
Add PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_{IMPORT,EXPORT} to prj.conf. This is
required by NCS to make mbedtls_pk_parse_key() work. In vanilla Zephyr
PSA_WANT_KEY_TYPE_ECC_KEY_PAIR_GENERATE was enough. Set both explicitly, so
that we are fine with future releases of Zephyr and mbedTLS.

Select PSA_WANT_RSA_KEY_SIZE_4096. Pouch's bundled CA (ISRG Root X1) is
RSA-4096, so under Oberon PSA the CA parse fails with -0x3B00 (
MBEDTLS_ERR_PK_INVALID_PUBKEY) unless this is enabled.

Select NCS specific PSA_WANT_GENERATE_RANDOM for http_client, which
otherwise fails with -0x0086 (PSA_ERROR_NOT_SUPPORTED) mbedTLS TLS
handshake error.

Measured peak main-thread usage (Zephyr's 'kernel thread stacks',
captured after a successful pouch sync):

```
  frdm_rw612        vanilla Zephyr  http_client:  3512 / 4096  85%
  frdm_rw612        vanilla Zephyr  coap_client:  3032 / 4096  74%
  nrf52840dk W5500  vanilla Zephyr  coap_client:  2680 / 4096  65%
  nrf52840dk W5500  NCS 3.4         coap_client:  7612 / 8192  93%
  nrf52840dk W5500  NCS 3.4         http_client:  7628 / 8192  93%
```

Vanilla Zephyr's tf-psa-crypto peaks at ~3.5 KiB during mbedtls's RSA-4096
verify of the ISRG Root X1 CA. NCS's Oberon PSA needs ~7.6 KiB for the same
operation because of larger intermediate buffers. 8 KiB fits both platforms
comfortably and keeping a single value in prj.conf is worth the ~4 KiB of
RAM on the vanilla-Zephyr targets.

Provide nRF9160dk overlay files and add support for 'modem' as a network
medium next to 'wifi'.


hasheddan (approved):
Thanks for working through this @mniestroj!

PR #275, opened 2026-07-10
CI: Add OTA test harness and use it in ESP-IDF http_client example

szczys:
- Add a platform-agnostic harness for working with OTA packages, cohorts, and release.
- Add an OTA test to the ESP-IDF http_client pytest
- Add the OTA-specific pytest flags to the HIL workflow

resolves: https://github.com/golioth/firmware-issue-tracker/issues/1047

### DIFF
--- a/.github/workflows/test-esp-idf-hil.yml
+++ b/.github/workflows/test-esp-idf-hil.yml
@@ -217,6 +217,9 @@ jobs:
             exit 1
           fi
 
+          # Ensure we can query the short hash
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+
           pytest -vv -rs -s                  \
             -c "$SAMPLE_DIR/pytest.ini"      \
             --rootdir "$SAMPLE_DIR"          \
@@ -225,6 +228,9 @@ jobs:
             --port "${!PORT_VAR}"            \
             --app-path "$SAMPLE_DIR"         \
             --build-dir build                \
+            --fw-update-bin "$SAMPLE_DIR/build/${{ matrix.sample.name }}_${{ inputs.idf_target }}_fwupdate.bin" \
+            --fw-update-ver "$(git rev-parse --short HEAD)" \
+            --fw-update-pkg-name "hil_ota_$IDF_TARGET"      \
             --erase-all n                    \
             --api-url "$GOLIOTH_API_URL"     \
             --api-key "$GOLIOTH_API_KEY"     \
--- a/.github/workflows/test-esp-idf-hil.yml
+++ b/.github/workflows/test-esp-idf-hil.yml
@@ -75,8 +75,16 @@ jobs:
           # Install Pouch dependencies
           pip install -r requirements-ci-esp-idf.txt --require-hashes
 
-          # Enable shorter 10s sync delay for Hil testing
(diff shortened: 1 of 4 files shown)


scripts/pytest-pouch/pytest_pouch/ota_harness.py
```diff
@@ -0,0 +1,140 @@
...
+            fw_update_bin,
+            fw_update_ver,
+            pouch_ota_package,
+        )
+        artifact = await project.artifacts.upload(
+            path=fw_update_bin,
+            version=fw_update_ver,
+            package=pouch_ota_package,
+        )
+        artifacts_to_cleanup.append(artifact.id)
```
szczys, line 132: We're not adding the artifact.id to the cleanup list if the artifact already existed on the server. This workflow will only clean up what it created.

I have another branch in progress that cleans up any orphans that are more than a day old (same approach we use for the Firmware SDK CI).

.github/workflows/test-esp-idf-hil.yml
```diff
@@ -75,8 +75,19 @@ jobs:
           # Install Pouch dependencies
           pip install -r requirements-ci-esp-idf.txt --require-hashes
 
-          # Enable shorter 10s sync delay for Hil testing
-          export SDKCONFIG_DEFAULTS="sdkconfig.defaults;pytest/sdkconfig.pytest.defaults"
+          # Ensure we can query the short hash
+          git config --global --add safe.directory $GITHUB_WORKSPACE
```
hasheddan: Looks like we only use the short hash in the rebuild step, should we move this there?

.github/workflows/test-esp-idf-hil.yml
```diff
@@ -93,6 +104,32 @@ jobs:
 
           tar -cf build-artifacts.tar "${artifact_files[@]}"
 
+      - name: Build update version of firmware
```
hasheddan, line 104: Will this run for all samples (not just OTA tests)?

szczys: This only runs for samples that have pytest set up (currently that's http_client). I plan to add the same tests for ble_gatt; that sample is ready for the same tests but I need to do some plumbing work to set up a gateway as part of the HIL test.

So currently 100% of the tests need this update binary and when ble_gatt is added to the tests it will need it as well.

hasheddan: Makes sense -- thanks for the clarification!

hasheddan (approved):

mniestroj (approved):

.github/workflows/test-esp-idf-hil.yml
szczys: tl;dr: We're still in agreement, but I wanted to clarify an error I made before.

I was mistaken, we actually do have pytest set up for ble_gatt, but it is not run in CI right now because that sample is not part of the matrix (same reason as stated above: no gateway plumbing yet). Since the tests are common to the sample harness it is effectively the same pytest running for both of these samples, although the OTA test isn't selected until that harness is added to sample. This will happen for the ble_gatt sample soon so that we have full sample test coverage on all esp-idf example applications.

PR #276, opened 2026-07-13
ci: add scheduled project cleanup workflow

szczys:
Add clean_ci_project script that removes old devices, artifacts, cohorts, releases, tags, and certificates older than 1 day. Artifact cleanup is scoped to packages starting with 'hil_ota_' to avoid removing manually-uploaded or non-HIL artifacts.

Scheduled daily at 11 PM UTC against PROD project.

This is functionally the same cleanup script as the Golioth Firmware SDK with artifact cleanup added and style/formatting fixed. Because the two are redundant, I propose we run the clean up in this repository and deprecate the cleanup workflow in the Firmware SDK.

PR to remove cleaning workflow from Firmware SDK repo: https://github.com/golioth/golioth-firmware-sdk/pull/918

### DIFF
--- a/.github/workflows/ci-clean-project.yml
+++ b/.github/workflows/ci-clean-project.yml
@@ -1,0 +1,36 @@
+name: "CI: Clean Project"
+
+on:
+  schedule:
+    # Run workflow at the start of every day (11 PM UTC)
+    - cron: "0 23 * * *"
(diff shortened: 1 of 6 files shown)


hasheddan (approved):
Thanks @szczys!

mniestroj (approved):

PR #279, opened 2026-07-14
Bluetooth Gateway: fix ENOMEM failures during block download

szczys:
This PR addresses issues with block-download on the bluetooth transport observed with both Zephyr and ESP-IDF leaf devices.

- port: use recursive mutex for FreeRTOS
- gateway: use work queue for pouch_gateway_uplink_close
- gateway: use timepoint for blockbuffer_alloc and pouch_msgq_put
- broker: add per-device lock to guard sender struct

The key resolution is to move pouch_gateway_uplink_close off of the transport thread so that CoAP operations don't block. In the case of the gatt transport, ACK messages were blocked by the CoAP operations. This change introduces a concurrency issue that is resolved by adding a lock for each device.

Also observed is the block download outpacing the blocks consumed by the transport. A timepoint is used to give both block alloc and message queue time to become free.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/1048


tests/pouch/gateway/CMakeLists.txt
```diff
@@ -34,5 +34,7 @@ target_compile_definitions(app PRIVATE
   CONFIG_POUCH_GATEWAY_NUM_BLOCKS=4
   CONFIG_POUCH_GATEWAY_DOWNLINK_BLOCK_TIMEOUT=1
   CONFIG_POUCH_GATEWAY_CLOUD=1
+  CONFIG_POUCH_GATEWAY_WORKQ_STACK_SIZE=2048
+  CONFIG_POUCH_GATEWAY_WORKQ_PRIORITY=5
```
mniestroj, line 38: Why do we change defaults here? Can we rely on Kconfig default instead?

szczys: I don't think the gateway Kconfig gets sourced during this test.  The comment at the top of this function:

```
# Provide the macros that the bypassed POUCH_GATEWAY Kconfig tree would
# normally generate.
```

edit: sorry, this answer was left as "pending", thanks for the bump.

szczys:
This initially failed the gateway build on the nrf9160dk because we ran out of flash. I've added commits to mitigate that:

- CONFIG_MCUMGR_TRANSPORT_WORKQUEUE_STACK_SIZE was set to 4096 in the prj.conf but the nrf52 doesn't need it (that fix was for the rw612
- Moving CONFIG_BT_MAX_CONN from 16 to 8 on this board frees up ~20k in RAM. This seems like a good choice since we were above 98% utilization before this change. I noticed heap issues with that much utilization so this gives us needed head room.

### Known issues

The block download completes successfuly with the nrf9160dk as a gateway, however we see these warnings for the blocks. I think this should be addressed in a different PR:

```
5.849,334] <wrn> modem_cmux: DLCI 2 receive buffer low, set flow control
[00:35:56.650,970] <wrn> modem_cmux: DLCI 2 receive buffer low, set flow control
[00:35:57.367,156] <wrn> modem_cmux: DLCI 2 receive buffer low, set flow control
[00:35:58.046,844] <wrn> modem_cmux: DLCI 2 receive buffer low, set flow control
[00:35:58.883,270] <wrn> modem_cmux: DLCI 2 receive buffer low, set flow control
[00:35:59.671,203] <wrn> modem_cmux: DLCI 2 receive buffer low, set flow control
[00:36:00.471,191] <wrn> modem_cmux: DLCI 2 receive buffer low, set flow control
[00:36:01.271,148] <wrn> modem_cmux: DLCI 2 receive buffer low, set flow control
```

mniestroj (approved):

tests/pouch/gateway/CMakeLists.txt
mniestroj: @szczys Bump on this one.

PR #285, opened 2026-07-16
ESP-IDF: Deduplicate delayable work queue and fix timer leak

szczys:
- Add use a QUEUED flag to ensure delayable work cannot be enqueued more than once
- Use the QUEUED flag to cancel delayable work already in the queue but not yet processed
- Reuse the delayalbe work timer when initialize is called more than once. This fixes leak that dropped the previous timer handle


mniestroj (approved):

trond-snekvik (approved):

port/include/pouch/port.h
```diff
@@ -791,8 +791,20 @@ int pouch_work_reschedule(pouch_work_delayable_t *dwork, pouch_timeout_t delay);
 /**
  * @brief Cancel a pending delayable work item
  *
- * Cancels any pending scheduled execution of the work item. If the work is currently
- * executing, this function returns immediately without waiting for it to complete.
+ * Cancels any pending scheduled execution of the work item. This is an
```
hasheddan, line 794: Thanks for the expanded context here!

hasheddan (approved):

PR #286, opened 2026-07-17
Replace encryption_none with encryption_mock

trond-snekvik:
Adds CONFIG_ZTEST as a requirement for running without encryption, and renames this mode to mock encryption to reinforce the fact that this is not a valid encryption scheme on target. Removes this option on esp-idf, where it's not in use.

Supersedes #186.


trond-snekvik:
This is a somewhat different approach from #186 that allows us to keep running tests like the downlink test in CI without allowing this unsafe mode in actual deployments. I think we would have ended up with something like this eventually anyway, and the only downside (imo) is the test code leaking into the API in the pouch_config. If we end up flattening the header to remove this mode from the spec altogether later, we can revisit this.

port/zephyr/Kconfig
hasheddan: Because we don't actually want consumers of Pouch to ever configure this, perhaps it would make sense to remove the `choice`, then make `POUCH_ENCRYPTION_SAEAD` default `y` and make both it and `POUCH_ENCRYPTION_MOCK` hidden. It would potentially help eliminate any confusion and would effectively reduce the kconfig API surface area, which we could always opt to expand it in the future if necessary.

trond-snekvik: Works for me. Removed `choice` and the prompt on SAEAD, and made it `default y if !POUCH_ENCRYPTION_MOCK`. We still need the prompt for the mock encryption to be able to select it in prj.conf in the tests, but it'll be unavailable and hidden for other configurations.

hasheddan (approved):

PR #291, opened 2026-07-20
CI: ESP-IDF: add ble_gatt to HIL testing

szczys:
- Add dependency on gateway build
- Flash gateway firmware to frdm_rw612 as part of ble_gatt HIL test
- Add gateway provisioning harness
- Update credential generation to share CA root cert for both gateway and device credential generation
- Add ble_gatt to the HIL testing matrix
- Update the connection confirmation string to work for all examples (previously http_client specific)

### Known issues

The build steps run in series which takes time:

    gateway-build -> hil-matrix -> device-build -> hil-test

This can be resolved but it requires breaking the last three jobs above in to two separate workflows. I'll do that in a future PR so this PR doesn't grow any larger.

resolves: https://github.com/golioth/firmware-issue-tracker/issues/975


.github/workflows/test-esp-idf-hil.yml
```diff
@@ -39,26 +39,17 @@ on:
...
   build:
     permissions:
       contents: read
       actions: write
     runs-on: ubuntu-24.04
-    needs: prepare-matrix
     strategy:
-      matrix: ${{ fromJson(needs.prepare-matrix.outputs.matrix) }}
+      fail-fast: false
+      matrix: &hil-matrix
```
szczys, line 49: I really love the change from json to yaml here, but alas this approach cannot be passed between workflows so when we break this workflow into two files for better parallel execution we're going to back to json.

requirements.txt
```diff
@@ -1,4 +1,4 @@
 cbor2<6.0.0
-typer<0.24.0
-click<8.2.0
+typer>=0.24.0,<0.25.0
```
trond-snekvik: Could we just lock this to 0.24?

trond-snekvik (approved):

PR #294, opened 2026-07-21
zephyr: add LED control the coap_client and http_client

szczys:
switch led0 on and off based on the LED setting. This maps to the same approach taken by ble_gatt.


szczys:
This needs a different approach so it doesn't break with native_sim. I'm going to close this PR and will track the future work here: https://github.com/golioth/firmware-issue-tracker/issues/1052

PR #298, opened 2026-07-22
Remove ignored Kconfig values in the samples

trond-snekvik:
A few of our settings in the samples have no effect, but generate build warnings. This removes those settings without affecting the builds.

- The `CONFIG_NET_CONNECTION_MANAGER_CONNECTIVITY_WIFI_MGMT` setting has missing dependencies, and don't have an effect.
- The `CONFIG_HEAP_MEM_POOL_SIZE` gets overwritten by the build system's kernel mem pool size calculation.
- The `CONFIG_MBED_HEAP_SIZE` reports missing dependencies, but it might be valid in NCS?


trond-snekvik:
I'm really posting this to start a discussion, as I expect I'm just missing something. Do the heap size settings have an effect on any of our supported platforms? 

I think the `CONFIG_NET_CONNECTION_MANAGER_CONNECTIVITY_WIFI_MGMT` change is valid, as it's only applied on the RW612, but I haven't tested all hardware/SDK combinations of the others. If they're required, I think we might want to push them to board specific files instead - although that might not be feasible if NCS makes a difference

szczys (approved):
Okay,  I went on kind of a deep dive over the removal of the `CONFIG_HEAP_MEM_POOL_SIZE` symbols. As @trond-snekvik already knows, I found out these affect the kernel heap size, which is constrained. The symbol does not affect pouch dynamic allocations because we do not use `k_malloc()` anywhere. Our `malloc()` calls allocate from whatever memory is leftover after static allocation (the non-kernel heap).

All other symbol removal proposed here looks fine as well. Thanks for the work on this!!

examples/zephyr/ble_gatt/prj.conf
```diff
@@ -31,8 +31,6 @@ CONFIG_BT_RX_STACK_SIZE=4096
 
 CONFIG_MAIN_STACK_SIZE=4096
 
-CONFIG_MBEDTLS_HEAP_SIZE=9192
```
szczys, line 34:
```suggestion
CONFIG_MBEDTLS_HEAP_SIZE=9192
```

I found in testing that his breaks the handshake:

```
[00:00:03.088,348] <err> cert: Failed to parse certificate: 0x8d
[00:00:03.088,378] <err> cert: Failed loading server cert
[00:00:05.082,794] <err> saead_session: Missing server key
[00:00:05.082,794] <err> saead_uplink: Session key generation failed
[00:00:27.888,793] <err> cert: Failed to parse certificate: 0x8d
[00:00:27.888,793] <err> cert: Failed loading server cert
[00:00:28.983,459] <err> pouch_sender: Error from endpoint, aborting
[00:00:51.853,271] <err> cert: Failed to parse certificate: 0x8d
[00:00:51.853,271] <err> cert: Failed loading server cert
```

szczys (changes requested):

examples/zephyr/ble_gatt/prj.conf
trond-snekvik: Reintroduced this symbol. It's only really required for NCS, so we could put it in a kconfig file somewhere to get out of the warning when building for zephyr, but it's application specific, so it's a bit hard to settle on a universal value.

hasheddan (approved):

mniestroj (approved):

szczys (approved):

Issue #274, opened 2026-07-10
RFC: rpmsg transport for Pouch on MPU+MCU systems


beriberikix:
# RFC: rpmsg transport for Pouch on MPU+MCU systems

> [!NOTE]
> Implemented and validated on hardware. The work has moved out of this repository and continues as a single PR on a fork: **beriberikix/pouch#1**.

## Proposal

Many SoCs pair a Linux core with a microcontroller core (NXP i.MX 8M/9, ST STM32MP, TI Sitara, Renesas RZ/G) and connect the two with OpenAMP/rpmsg. This proposes a Pouch transport over that link. The MCU runs the Pouch device stack with its own identity and end-to-end encryption, and a daemon on the Linux side forwards opaque pouches to Golioth without ever holding the device's keys.

Because the MCU usually has no flash of its own, the Linux side also applies its firmware updates, by installing a new image and restarting the core through remoteproc.

## Outcome

- The transport became an adapter on the existing serial transport core, using a single rpmsg endpoint and the resource table.
- Firmware reaches the MCU two ways, both optional: relayed through the MCU from Pouch OTA, or fetched directly by the Linux side from a URL the MCU signs.
- Validated on an FRDM-IMX95 against production Golioth, including unattended OTA updates through both paths.

## Still open

- Linux-mediated OTA without a per-download signature, which would take the device out of the transfer path entirely.
- Where the gateway daemon should live. We expect it to end up in a production gateway rather than in this repository.

beriberikix:
This work has moved out of the repository. #304, #349, #350 and #351 are closed and consolidated into a single PR on a fork: beriberikix/pouch#1.

That PR carries the signed-URL delivery this comment used to describe, and the review feedback from #304 verbatim as open items. The detailed findings, including the server's zero-tolerance handling of a future not-before timestamp, remain in the branch's commit messages.

PR #256, opened 2026-07-02
transport: serial: add SPI transport adapter

trond-snekvik:
Add SPI master (broker) and slave (device) adapters for the pouch serial transport, including devicetree bindings, Kconfig options, and example applications for both sides.

Rework the serial channel state machine to support explicit open/close transitions, channel suspension, and per-channel active tracking in the broker. The broker now restarts the exchange loop on completion or error instead of relying on the caller.

Also includes:
- Gateway cert logging and NULL-checks.
- dts_root entry in module.yml so bindings are picked up.

Targets the transport/serial branch, so staying in draft for now

PR #304, opened 2026-07-23
transport: rpmsg (OpenAMP) transport for MPU+MCU systems

beriberikix:
Device-side rpmsg transport from the RFC in #274: Pouch on the MCU core of a heterogeneous MPU+MCU SoC, syncing to a broker on the Linux core over on-die OpenAMP/rpmsg. The MCU keeps its own cloud identity and end-to-end encryption; the Linux side only ever handles ciphertext.

Per the RFC this is an **adapter on the serial transport core (#232)**, not a new transport — one rpmsg message carries one Pouch Serial frame, so there is no length prefix and no SAR.

## Reviewing this

Four commits, each self-contained; worth reading in order rather than as a single diff.

| Commit | What to look at |
|---|---|
| `transport: serial: add rpmsg and UART device adapters` | The bulk of the review: the resource-table rpmsg adapter and the UART one. |
| `tests: pouch: cover the rpmsg/UART framing…` | Framing round-trip, resync, malformed/oversize rejection, and a full exchange routed through it. |
| `examples: zephyr: add the rpmsg_device sample` | Builds two ways from one tree; the board conf documents non-obvious tuning. |
| `ci: build the rpmsg example for imx95_evk` | Small but load-bearing — see below. |

**Why the resource table, not `ipc_service`.** Zephyr's DT-instantiable `ipc_service` backends use a static vring layout that the Linux kernel's `virtio_rpmsg_bus` does not understand, so against Linux the vdev has to come from the **resource table**. That is what `rpmsg_rsc_device.c` does. An earlier revision of this branch also carried an `ipc_service` adapter for Zephyr-to-Zephyr peers; it was dropped before review, because nothing in the tree enabled it, no build compiled it, and there was no use case asking for it.

**The CI commit matters more than its size suggests.** Nothing compiled either rpmsg adapter before it: the sample's only allowed platforms were `native_sim`, which selects the UART adapter, so `rpmsg_rsc_device.c` could break unnoticed and the `open-amp`/`libmetal` manifest entries bought nothing. Adding the board to `platform_allow` would not have helped — the nsim job passes an explicit `-p native_sim` list, so a build-matrix entry is the only route.

## Two fixes only a fast host exposes

- `rpmsg_send()` blocks until a TX buffer frees, and it is the thread servicing the vrings that would free one. Uses `rpmsg_trysend()` with a bounded retry.
- The management thread ran at the same priority as the Pouch work queue, so it drained the entire receive vring without yielding while the decrypt queue never ran. A host pushed 137 KB in 6 ms and `pouch_downlink_push()` kept heap-allocating encrypted blocks until `malloc` failed, leaving the downlink channel permanently erroring. It now runs below the work queue.

## Hardware validation

FRDM-IMX95 (i.MX95 15x15), Cortex-M7 under Linux `remoteproc`, against **production Golioth**, brokered by the Linux gateway in the stacked branch below.

- The M7 boots under `remoteproc` and brings the adapter up; the endpoint auto-binds to `rpmsg_char` as `/dev/rpmsg0` with no `RPMSG_CREATE_EPT_IOCTL`.
- With production credentials `pouch_init()` succeeds and the DEVICE_CERT channel serves the certificate byte-identical (385 bytes).
- A full session runs — INFO → SERVER_CERT → DEVICE_CERT → UPLINK → `POST /.g/pouch` → DOWNLINK — the device registers in the project, and telemetry lands in LightDB.
- A second session skips both certificate phases, because the device reports back the server-certificate serial it now holds.
- The transport was also driven under load by the stacked work: a 250 KB firmware image relayed through it in chunks, arriving byte-identical to the cloud artifact by SHA-256.

## The stack

1. **this PR** — the device-side transport.
2. `transport/rpmsg-broker` — the Linux gateway (Go) that brokers to the cloud; this is what made the validation above possible. Its home is still open, since connect-gateway may absorb it.
3. `transport/rpmsg-ota` — MCU-mediated OTA (a core with no flash relays its image to the host, which verifies and applies it via remoteproc). Gated `default n`, and it carries a fix for a NULL-endpoint dereference the extra channels would otherwise expose.

Split out of one larger PR so the transport can be reviewed on its own: the OTA half mutates the shared `pouch_serial_channel_id` enum that every serial device inherits, which RFC #274's open question 2 flagged as debatable, and that discussion should not hold up the transport.

## Notes

- The serial-core fixes once split into #317 are already in `main`, folded into #232 before it merged.
- The example's device credentials are self-signed placeholders and the README says so; validation used real ones kept out of the tree.
- Occasional `Gateway Build` failures on this branch are GitHub's artifact service returning 403 on upload, not build errors.


examples/zephyr/rpmsg_device/Kconfig
trond-snekvik, line 1: Prefer to not include an application Kconfig file when there are no application specific symbols.

port/zephyr/transport/serial/uart_device.c
```diff
@@ -0,0 +1,181 @@
...
+        int n = uart_fifo_read(dev, buf, sizeof(buf));
+        if (n <= 0)
+        {
+            break;
+        }
+
+        uint32_t written = ring_buf_put(&rx_ring, buf, n);
+        if (written < (uint32_t) n)
+        {
+            LOG_WRN("RX ring overflow, dropped %d bytes", n - (int) written);
```
trond-snekvik, line 68: This is concerning. If this happens, we can drop a few bytes in the middle of a frame without detecting it, and end up passing one corrupted frame, and then drop the next one, as we ate the 0xA5 and length in the corrupted frame.

Instead of always reading 64 bytes from the uart fifo into a local buffer, then figure out whether we can fit it, we could use `ring_buf_put_claim` and `ring_buf_put_finish` here to only move the data we can fit:

```c
uint8_t *data = NULL;
uint32_t maxlen =
	ring_buf_put_claim(&rx_ring, &data, CONFIG_POUCH_SERIAL_UART_DEVICE_RX_RING_SIZE);
if (maxlen == 0 || data == NULL) {
	break;
}
uint32_t written = uart_fifo_read(dev, data, maxlen);
ring_buf_put_finish(&rx_ring, written);
```

This is less copying, and we won't drop data in this location.

port/zephyr/transport/serial/uart_framing.c
```diff
@@ -0,0 +1,83 @@
...
+    framer->expected = 0;
+    framer->got = 0;
+}
+
+size_t pouch_uart_framer_feed(struct pouch_uart_framer *framer, uint8_t byte)
+{
+    switch (framer->state)
+    {
+        case POUCH_UART_FRAMER_SOF:
+            if (byte == POUCH_UART_FRAME_SOF)
```
trond-snekvik, line 45: What happens if we are in SOF mode after a corrupted frame, and encounter `0xA5` in the payload? And what happens if that's the last byte in the payload before the actual `0xA5` byte?

port/zephyr/transport/serial/uart_device.c
trond-snekvik, line 1: I think this should use the existing UART adapter queued up in https://github.com/golioth/pouch/tree/transport/uart. It's not in main yet (it's been stuck in review purgatory behind #256 since early July), but it resolves a few of the open issues in this adapter. We don't want two separate UART adapter implementations going forwards.

port/zephyr/transport/serial/rpmsg_rsc_device.c
```diff
@@ -0,0 +1,423 @@
...
+    {
+        /* Poll rather than block indefinitely. The serial core's frame_get()
+         * returns 0 when an endpoint reports POUCH_MORE_DATA with no bytes yet
+         * ("wait for the next call", channel.c), but nothing schedules that
+         * call: the uplink endpoint produces its data asynchronously on the
+         * pouch work queue and never invokes bearer_ready(). Waiting only on
+         * the semaphore would therefore stall a session forever in the window
+         * between the broker's prompt and the first encrypted block.
+         */
+        k_sem_take(&tx_sem, K_MSEC(CONFIG_POUCH_RPMSG_RSC_DEVICE_TX_POLL_MS));
```
trond-snekvik, line 380: If the problem in the comment here is true, this is something that should be addressed in the uplink endpoint, and not something we should be adding workarounds for in this adapter. Does `K_FOREVER` actually cause issues here?

examples/zephyr/rpmsg_device/src/credentials.c
```diff
@@ -0,0 +1,72 @@
...
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(credentials, LOG_LEVEL_INF);
+
+static const uint8_t device_key[] = {
+#include "device_key.der.inc"
+};
+
+static const uint8_t device_crt[] = {
+#include "device_crt.der.inc"
+};
```
trond-snekvik, line 29: Woah, this has to change before we merge. 

It's just a placeholder (presumably?) in example app, but examples are there to demonstrate how to use our stuff, and under no circumstance should we demonstrate the practice of committing private keys to source control and baking them into the application binary, even if it's a placeholder. 

The other samples do this with mcumgr, and I suggest we do the same here. If this isn't an option for this platform, we need to discuss some alternatives.

trond-snekvik (changes requested):
To be frank, I'm not entirely sure what to do with this PR. Aside from some duplication with other pending development work and some other open issues, the code doesn't _look_ bad, but AI code rarely does on the surface. It is however nearly 2000 lines of code generated by Claude and submitted without a human developer allocated to actually maintain and reason about it. We need to have an internal discussion on what sort of workflow we want to have in this repo moving forwards, because at this rate, it's becoming unreasonable to expect the dev team to be able to maintain it.

beriberikix:
Moving this out of the main repository. It continues as a single PR on my fork, beriberikix/pouch#1, which also consolidates #349, #350 and #351. The review feedback above is carried over there verbatim as open items. Closing here.

PR #306, opened 2026-07-23
Thingy91x Gateway Fixes

szczys:
- Update gateway_custom_connect partitions to match changes already made in gateway so code is generated without a bootloader offset
- Make sure UART0 is disabled on ncs-serial-modem as it inteferes with inter-chip comms.
- Handle memory use in the same way as the nrf9160dk so we have enough room for heap
- Add auto-pairing by default as is done wiht nrf9160dk
- Document how to provision certificates by flashing an lfs binary

### Known Issues:

The chips can now communicate with one another and we are able to get a network connection and start the sync with bluetooth devices. However, after every pouch operation, the BLE connection drops with `reason 0x08`
~5-8 seconds later, regardless of the 7-second supervision timeout.

This issue is beyond the scope of the current PR and is tracked here: https://github.com/golioth/firmware-issue-tracker/issues/1056


examples/zephyr/gateway_custom_connect/README.md
```diff
@@ -53,6 +53,84 @@ mcumgr --conntype serial --connstring $SERIAL_PORT fs upload $KEY_FILE /lfs1/cre
...
+storage partition on the device.
+
+This approach is useful when MCUmgr USB/serial provisioning is not available,
+such as on platforms where the UART is used for other purposes (e.g. serial
+modem communication on a Thingy91x-based gateway).
+
+Install the `littlefs_tools` Python package:
+
+```bash
+pip install littlefs_tools
```
trond-snekvik, line 69: Could potentially go into requirements.txt or a new examples/requirements.txt

trond-snekvik (approved):
Not really for this PR, but it'd be real neat if we made this a west command - something like `west provision <crt> <privatekey>`. I know the west runner mechanism should allow us to interact with whatever flash command the hardware needs, but I have never tried it. 

Could be something we could throw to an LLM and see if it manages to get it right

examples/zephyr/gateway/Kconfig
```diff
@@ -15,12 +15,14 @@ config SAMPLE_POUCH_GATEWAY_BT_AUTO_BOND
 	bool
 	default y if BOARD_NRF52_BSIM_NATIVE
 	default y if BOARD_NRF9160DK_NRF52840
+	default y if BOARD_THINGY91X_NRF5340_CPUAPP_NS
```
mniestroj, line 18: I am not sure we need that for Thingy91:X. Did you try to use the button for pairing? I think it worked before. nrf9160dk has no button dedicated to nrf52840 core, so `default y if BOARD_NRF9160DK_NRF52840` was just a workaround for that.

szczys: Line 13 comments on this symbol and says:
```
# This option is used only for automated tests
config SAMPLE_POUCH_GATEWAY_BT_AUTO_BOND
	bool
	default y if BOARD_NRF52_BSIM_NATIVE
	default y if BOARD_NRF9160DK_NRF52840
	default y if BOARD_THINGY91X_NRF5340_CPUAPP_NS
```

So I added this to ensure we can run automated testing without pushing the button to pair.

examples/zephyr/gateway_custom_connect/README.md
szczys: Great, I added this to requirements.txt and regenerated requirements-ci-zephyr.txt.

examples/zephyr/gateway/Kconfig
mniestroj: Makes sense then :)

mniestroj (approved):

PR #317, opened 2026-08-10
fix(serial): guard optional endpoint->end and re-arm a stalled channel

beriberikix:
Two bugs in the serial channel state machine, both found while running an rpmsg transport built on this branch against a real broker on hardware (i.MX95 Cortex-M7 under Linux remoteproc, connect-agent as the broker). Both were reported as review comments on #232 — [here](https://github.com/golioth/pouch/pull/232#discussion_r3695491746) and [here](https://github.com/golioth/pouch/pull/232#discussion_r3695492089) — this is the patch for them.

### 1. `pouch_serial_ch_close()` dereferences an optional callback

`ch->endpoint->end()` is called unconditionally, but `end` is optional in the same way it is for the SAR sender/receiver: endpoints with nothing to tear down, such as the device info endpoint, leave it `NULL`.

That means a jump through a NULL pointer on the **first ACK any broker sends** — so every device crashes as soon as it is talked to.

### 2. `pouch_serial_ch_frame_get()` drops the channel's pending flag

`CH_FLAG_PENDING` is consumed on entry, but when the endpoint reports `POUCH_MORE_DATA` with zero bytes — "nothing yet, call me again" — the function returns without restoring it, so the channel is never polled again.

The device uplink endpoint hits this every session: it produces its data asynchronously on the Pouch work queue a few hundred milliseconds after the broker's prompt, and never calls `bearer_ready()`. The uplink transfer then stalls until the broker times the session out.

### Why CI doesn't catch either

Both are invisible to the current tests. Every mock in `stub_endpoints.c` defines `.end` (20 of 20 vtables), so #1 never fires; and they are all synchronous, so none of them ever returns `MORE_DATA` with no bytes and #2 never fires. Adding one endpoint without `.end`, and one that defers its first block, would cover both.

### DIFF
--- a/include/pouch/transport/serial/broker.h
+++ b/include/pouch/transport/serial/broker.h
@@ -1,0 +1,143 @@
+/*
+ * Copyright (c) 2026 Golioth, Inc.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include <pouch/port.h>
+
+/**
+ * @file broker.h
+ * @brief Pouch Serial transport (broker side)
+ *
+ * The broker is the master of the serial exchange. It initiates all transfers,
+ * drives the provisioning and data exchange sequence, and owns a
+ * @ref pouch_serial_broker instance allocated by the transport adapter.
+ *
+ * Typical usage:
+ *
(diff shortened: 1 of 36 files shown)


trond-snekvik (approved):

trond-snekvik:
Instead of working through the test failures here, I incorporated this directly into #232.

PR #319, opened 2026-08-13
Add pytest support for frdm_rw612 HIL testing

szczys:
This is preliminary work for implementing HIL testing for the Zephyr samples:
- Refactor native_sim harnesses to prepare for HIL testing
- Add pytest support for rw612 (ethernet) to the http_client example

resolves https://github.com/golioth/firmware-issue-tracker/issues/1058


examples/zephyr/native_serial/mcumgr_echo/pytest/conftest.py
```diff
@@ -15,6 +16,12 @@
 
 from twister_harness.device.device_adapter import DeviceAdapter
 
+sys.path.insert(
+    0, str(Path(__file__).resolve().parents[5] / "scripts" / "pytest-pouch")
+)
+
+pytest_plugins = ["pytest_pouch.plugin"]
+
```
szczys, line 24: These three tests (mcumgr_echo, mcumgr_shell, and serial_comm) don't actually use the plugin, but they need to include it so that pytest command-line flags passed in the twister call don't throw an error. For instance `--generate-certs` and `--ota-dummy-binary`

scripts/pytest-pouch/pytest_pouch/ota_harness.py
```diff
@@ -97,45 +81,86 @@ async def ota_update(
...
-
-    if matching:
-        artifact = matching[0]
-        logging.info(
-            "Found existing artifact (id=%s) for package=%s, version=%s — skipping upload",
-            artifact.id,
-            pouch_ota_package,
-            fw_update_ver,
+    fw_bin_path = request.config.getoption("--fw-update-bin")
+    fw_ver = request.config.getoption("--fw-update-ver")
```
trond-snekvik: Should this request the fixtures instead? We'll no longer fail/skip if the arg is missing

examples/zephyr/coap_client/pytest/conftest.py
```diff
@@ -0,0 +1,25 @@
...
+pytest_plugins = [
+    "pytest_pouch.plugin",
+    "pytest_pouch.ota_harness",
+    "pytest_pouch.zephyr_harness",
+]
+
+
+@pytest.fixture(scope="module")
+def ensured_credentials(creds):
+    pass
```
trond-snekvik: Potential noob question: is this required when the `zephyr_settings_tests` has one that gets imported in test_sample.py?

examples/zephyr/http_client/sample.yaml
```diff
@@ -4,18 +4,15 @@ common:
   harness: pytest
   harness_config:
     pytest_dut_scope: module
-  platform_allow:
-    - native_sim
-    - native_sim/native/64
-  integration_platforms:
```
trond-snekvik, line 10: `integration_platform` powers twister in the zephyr CI, do we just run every `platform_allow` our setup, or do we lose anything by removing it here (without adding it to the test cases)?

scripts/pytest-pouch/pytest_pouch/ota_harness.py
szczys: Ah,  I see what you mean. I've used the fixture with a slight update to handle the --ota-dummy-binary case.

examples/zephyr/coap_client/pytest/conftest.py
szczys: Ah, good catch. This adds the creds dependency which I originally used because of an issue between device testing and native_sim testing. But that issue was later resolved and this should have been removed. The fixtures that depended on `ensured_credentials` now depend directly on `creds`. Thanks!

examples/zephyr/http_client/sample.yaml
szczys: With this in place we log errors on the ` pouch.http_client.hil` test because the we don't run native_sim for those. You're correct that we fall back to the platform_allow with this removed.

trond-snekvik (approved):

hasheddan (approved):

PR #321, opened 2026-08-14
CI: add stream to Zephyr sample tests

szczys:
relates to https://github.com/golioth/firmware-issue-tracker/issues/1060

### DIFF
--- a/scripts/pytest-pouch/pytest_pouch/zephyr_stream_tests.py
+++ b/scripts/pytest-pouch/pytest_pouch/zephyr_stream_tests.py
@@ -1,0 +1,49 @@
+#
+# Copyright (c) 2026 Golioth, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+import datetime
+
+import anyio
+import pytest
+
+pytestmark = pytest.mark.anyio
+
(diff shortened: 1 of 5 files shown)


scripts/pytest-pouch/pytest_pouch/zephyr_stream_tests.py
```diff
@@ -0,0 +1,49 @@
...
+
+_STREAM_POLL_ATTEMPTS = 24
+_STREAM_POLL_DELAY_S = 5.0
+
+
+def _iso8601_utc(timestamp: datetime.datetime) -> str:
+    return timestamp.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+
+
+async def test_sensor_stream_uplink(device):
```
mniestroj, line 22: Is that automatically executed for each test? This commit message doesn't say much, so not sure what and why we are doing.

szczys: This file is included by test_sample.py in the relevant example apps so that we test stream data arriving on the cloud as part of the CI run. I updated the commit message to reflect this.

mniestroj (approved):

PR #323, opened 2026-08-14
ci: ble sample test collects and uploads gateway logs

szczys:
I've observed flaky BLE tests and in each it looks like the device is never able to connect to the gateway for any of the  three tests. This PR adds visibility into this type of failure but:
1. Collecting logs from the gateway and uploading them as part of the device test run artifacts
2. Ensuring the gateway connects to the cloud before considering that pytest fixture to be successful

### DIFF
--- a/.github/workflows/test-esp-idf-hil.yml
+++ b/.github/workflows/test-esp-idf-hil.yml
@@ -309,6 +309,7 @@ jobs:
           path: |
             pytest-${{ matrix.name }}-hil.log
             ${{ env.UNITY_REPORT_NAME }}
+            gateway-${{ matrix.name }}.log
 
       - name: Erase flash
         if: always()
--- a/scripts/pytest-pouch/pytest_pouch/gateway_provision.py
+++ b/scripts/pytest-pouch/pytest_pouch/gateway_provision.py
@@ -7,9 +7,12 @@
 import logging
 import secrets
 import subprocess
+import threading
 import time
+from pathlib import Path
 
 import pytest
+import serial
 
 
 def pytest_addoption(parser):
@@ -114,6 +117,24 @@ def gateway_creds(creds, creds_dir, gateway_creds_dir, gateway_cloud_device, pro
     )
 
 
+_GW_CONNECTED_PATTERN = "Scanning successfully started"
+_GW_LOG_FILENAME = "gateway-ble_gatt.log"
+_GW_REBOOT_TIMEOUT_S = 120.0
+_GW_BAUD = 115200
+
+
+def _read_loop(ser, log_path, stop_event):
+    with open(log_path, "ab") as f:
+        while not stop_event.is_set():
+            try:
+                data = ser.read(4096)
+            except serial.SerialException:
+                break
+            if data:
+                f.write(data)
+                f.flush()
+
+
 @pytest.fixture(scope="module", autouse=True)
 def provisioned_gateway(gateway_serial_port, gateway_creds_dir, gateway_creds):
     logging.info("Uploading gateway credentials via smpmgr")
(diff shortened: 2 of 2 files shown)


scripts/pytest-pouch/pytest_pouch/gateway_provision.py
```diff
@@ -114,6 +117,24 @@ def gateway_creds(creds, creds_dir, gateway_creds_dir, gateway_cloud_device, pro
     )
 
 
+_GW_CONNECTED_PATTERN = "Scanning successfully started"
+_GW_LOG_FILENAME = "gateway-ble_gatt.log"
```
mniestroj, line 121: Could the artifact destination be exposed as a pytest option such as `--gateway-log-file` instead of embedding the CI matrix name in this plugin (`ble_gatt`)?

The workflow could then pass `gateway-${SAMPLE_NAME}.log`, making cleanup, capture, and upload use one explicit path.

.github/workflows/test-esp-idf-hil.yml
mniestroj, line 285: We can pass `$GATEWAY_ARGS` here, so that both serial port name and gateway log file is passed here.

scripts/pytest-pouch/pytest_pouch/gateway_provision.py
```diff
@@ -148,7 +169,57 @@ def provisioned_gateway(gateway_serial_port, gateway_creds_dir, gateway_creds):
...
+            pytest.fail(
+                f"Gateway failed to start scanning within {_GW_REBOOT_TIMEOUT_S}s. "
+                f"Last captured output: {last_content[-200:]}\n"
+                f"Full gateway log: {_gw_log}"
+            )
+    except Exception:
+        stop_event.set()
+        reader.join(timeout=5)
+        ser.close()
+        raise
```
mniestroj, line 218: `pytest.fail()` raises `_pytest.outcomes.Failed`, which derives from `BaseException`, not `Exception`. The timeout path therefore skips this handler and leaves the serial port and reader thread running. Could resource ownership be moved into a try/finally or context manager covering both the readiness wait and fixture yield?

scripts/pytest-pouch/pytest_pouch/gateway_provision.py
```diff
@@ -148,7 +169,57 @@ def provisioned_gateway(gateway_serial_port, gateway_creds_dir, gateway_creds):
...
+    reader.start()
+
+    try:
+        logging.info("Waiting for gateway connected pattern: %s", _GW_CONNECTED_PATTERN)
+
+        deadline = time.monotonic() + _GW_REBOOT_TIMEOUT_S
+        while time.monotonic() < deadline:
+            if _gw_log.exists():
+                try:
+                    content = _gw_log.read_text()
```
mniestroj, line 194: UART output is arbitrary bytes, but `read_text()` decodes it as strict UTF-8. A single invalid byte raises `UnicodeDecodeError,` and the broad handler below silently ignores it, so this can time out even when the readiness marker was received. This also rereads the complete growing file once per second while another thread writes it. Could the reader match a byte marker incrementally and signal readiness through an `Event` or `Queue` instead?

scripts/pytest-pouch/pytest_pouch/gateway_provision.py
```diff
@@ -114,6 +117,24 @@ def gateway_creds(creds, creds_dir, gateway_creds_dir, gateway_cloud_device, pro
...
+_GW_REBOOT_TIMEOUT_S = 120.0
+_GW_BAUD = 115200
+
+
+def _read_loop(ser, log_path, stop_event):
+    with open(log_path, "ab") as f:
+        while not stop_event.is_set():
+            try:
+                data = ser.read(4096)
+            except serial.SerialException:
```
mniestroj, line 131: Can this exception be propagated back to the fixture instead of silently breaking the reader loop? As written, a serial disconnect or read failure becomes a generic 120-second readiness timeout, hiding the root cause that these logs are intended to expose. A status queue could deliver either readiness or the original exception.

.github/workflows/test-esp-idf-hil.yml
```diff
@@ -309,6 +309,7 @@ jobs:
           path: |
             pytest-${{ matrix.name }}-hil.log
             ${{ env.UNITY_REPORT_NAME }}
+            gateway-${{ matrix.name }}.log
```
mniestroj, line 312: Please also remove this file in the Clean stale run assets step. This runs on a self-hosted runner, and the artifact step executes on failure. If the job fails before the fixture reaches _gw_log.unlink(), it can upload a gateway log left by an earlier run.

scripts/pytest-pouch/pytest_pouch/gateway_provision.py
```diff
@@ -7,9 +7,12 @@
 import logging
 import secrets
 import subprocess
+import threading
 import time
+from pathlib import Path
 
 import pytest
+import serial
```
mniestroj, line 15: Should we add `pyserial` to `requirements-ci-esp-idf.in` and/or plugin's `pyproject.toml`?

mniestroj:
Merged alternative PR: #326

PR #325, opened 2026-08-18
CI: Add fw_update to Zephyr HIL test

szczys:
Adds a fw_update test to the Zephyr HIL run. Currently frdm_rw612 on http_client is the only supported target but the implementation has been designed to scale.

1. Refactor pytest fw_update flags, removing --ota-dummy-binary in favor of `--ota-mode=[firmware | dummy | disabled]`
2. Update ESP-IDF harness and tests to use new flag approach
3. Add fw_update test to zephyr_ota_tests.py; pytest.mark.ota_mode is used to select between tests
4. Add workflow for fw_update HIL test
    1. Cloud update bin collision (component name and version) is avoided by leveraging the harness to generate a per-sample name.
    2. Use the short hash to generate a imgtool compatible version number
    3. Replace the VERSION files and rerun twister build step too generate update binaries
5. Use -vv instead of -v for better twister CI clarity
6. Revise pytest documentation for ESP-IDF to reflect repo changes and fw_update

resolves https://github.com/golioth/firmware-issue-tracker/issues/1060

### DIFF
--- a/examples/esp_idf/http_client/pytest/README.md
+++ b/examples/esp_idf/http_client/pytest/README.md
@@ -32,21 +32,6 @@ uv pip install --python "$IDF_PYTHON_ENV_PATH" "golioth@git+https://github.com/g
 
    - `examples/esp_idf/http_client/pytest/json-sensor-path-to-lightdb-stream.txt`
 
-## Build Firmware
-
-From the repository root (`pouch`):
-
-```bash
-idf.py -C examples/esp_idf/http_client set-target esp32s3
-export SDKCONFIG_DEFAULTS="sdkconfig.defaults;pytest/sdkconfig.pytest.defaults"
-idf.py -C examples/esp_idf/http_client build
-```
-
-This build command includes an overlay file that resets the sync
-interval to 10s. This may be omitted to use the default 30s period.
-
-## Run Test
-
 Set required environment variables (test auto-provisions before cloud
 checks):
 
@@ -55,28 +40,57 @@ export GOLIOTH_API_URL="https://api.golioth.io"
 export GOLIOTH_API_KEY="your_project_api_key"
 export WIFI_SSID="your_wifi_ssid"
 export WIFI_PSK="your_wifi_psk"
-
-# Manual cert mode: provide DER file paths
-export DEVICE_CRT_DER_PATH="/path/to/device.crt.der"
-export DEVICE_KEY_DER_PATH="/path/to/device.key.der"
-
-# Optional override (otherwise derived from cert CN in DEVICE_CRT_DER_PATH)
-export GOLIOTH_DEVICE_NAME="your_device_name"
 ```
 
-The test also accepts `python-golioth-tools`/SDK-style pytest flags:
+## Build Update Firmware
 
-- `--api-key` (fallback: `GOLIOTH_API_KEY`)
-- `--api-url` (fallback: `GOLIOTH_API_URL`, default: `https://api.golioth.io`)
-- `--device-name` (fallback: `GOLIOTH_DEVICE_NAME`)
-- `--wifi-ssid` (fallback: `WIFI_SSID`)
-- `--wifi-psk` (fallback: `WIFI_PSK`)
+Note: if you prefer to skip the fw_update test you can ignore this
+section and simply add the following to the pytest command in place of
+fw_update specific flags:
 
-To generate device certificates automatically during the test run,
-pass `--generate-certs` and omit manual cert inputs
-(`DEVICE_CRT_DER_PATH`/`DEVICE_KEY_DER_PATH`).
+```
+--ota-mode=disabled
+```
 
-Then run from the repository root (`pouch`):
+OTA testing requires two firmware builds: an update version (unique
+version number) and an initial version (default version).
+
+1. Build the update firmware with a unique version and OTA package name:
+
+```bash
+# Create fw_update.defaults with OTA package name and version
+cat > examples/esp_idf/http_client/fw_update.defaults <<EOF
+CONFIG_EXAMPLE_FW_UPDATE_COMPONENT="hil_ota_esp32s3_http_client"
+CONFIG_APP_PROJECT_VER="$(git rev-parse --short HEAD)"
+EOF
+
+idf.py -C examples/esp_idf/http_client set-target esp32s3
+export SDKCONFIG_DEFAULTS="sdkconfig.defaults;pytest/sdkconfig.pytest.defaults;fw_update.defaults"
+rm -f examples/esp_idf/http_client/sdkconfig
+idf.py -C examples/esp_idf/http_client reconfigure
+idf.py -C examples/esp_idf/http_client build
+
+# Copy the update binary to a known location
+cp examples/esp_idf/http_client/build/pouch_http_client_example.bin /tmp/http_client_esp32s3_fwupdate.bin
+```
+
+2. Build the initial version of the example:
+
+```bash
+# Create fw_update.defaults with OTA package name (no version override)
+cat > examples/esp_idf/http_client/fw_update.defaults <<EOF
+CONFIG_EXAMPLE_FW_UPDATE_COMPONENT="hil_ota_esp32s3_http_client"
+EOF
+
+export SDKCONFIG_DEFAULTS="sdkconfig.defaults;pytest/sdkconfig.pytest.defaults;fw_update.defaults"
+rm -f examples/esp_idf/http_client/sdkconfig
+idf.py -C examples/esp_idf/http_client reconfigure
+idf.py -C examples/esp_idf/http_client build
+```
+
+## Run Test
+
+From the repository root (`pouch`):
 
 ```bash
 # Full erase to remove existing credentials:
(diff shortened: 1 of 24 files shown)


scripts/pytest-pouch/pytest_pouch/zephyr_ota_tests.py
```diff
@@ -12,14 +12,18 @@
 
 pytestmark = pytest.mark.anyio
 
+OTA_DOWNLINK_TIMEOUT_S = 120.0
+OTA_DOWNLOAD_TIMEOUT_S = 180.0
```
szczys, line 17: Download time will almost certainly need to be configurable for coap_client test. I've left that for future work in favor of an easy constant value here.

scripts/pytest-pouch/pytest_pouch/zephyr_harness.py
```diff
@@ -18,6 +20,79 @@
...
+
+def _fw_update_bins_root(device_object) -> Path:
+    """Compute the fw_update_bins extraction root for the current test's build."""
+    app_build_dir = device_object.device_config.app_build_dir
+    parts = app_build_dir.parts
+    try:
+        idx = parts.index("twister-out")
+    except ValueError:
+        pytest.fail(f"Expected 'twister-out' in build path: {app_build_dir}")
+    return Path(*parts[:idx], "twister-out", "fw_update_bins")
```
szczys: This smells a little bit since it relies on the workflow to unpack update binaries in a very specific place. However, the approach uses the application build directory from the build step to match up with the application build directory of the update binary step which I think is itself a fairly safe (and hands-off) method of doing it.

scripts/pytest-pouch/pytest_pouch/plugin.py
```diff
@@ -47,13 +47,48 @@ def pytest_addoption(parser):
...
+            if marker is None:
+                keep.append(item)
+            else:
+                drop.append(item)
+        else:
+            if marker is None or marker.args[0] == selected:
+                keep.append(item)
+            else:
+                drop.append(item)
+    if drop:
```
mniestroj: A marker with no positional argument, including a kwargs-only declaration, reaches `marker.args[0]` in active `dummy` or `firmware` mode and aborts collection with `IndexError`. An unknown first value is silently deselected, while extra arguments and keyword arguments are ignored when the first positional value matches. With `--ota-mode=disabled`, every marked test is dropped without validation.

Malformed declarations can therefore crash collection, silently remove coverage, or be accepted depending on the selected mode. Please validate every non-null marker before both selection branches, require exactly one positional value (`dummy` or `firmware`) and no keyword arguments, and raise a node-specific `pytest.UsageError`.

scripts/pytest-pouch/pytest_pouch/gateway_provision.py
```diff
@@ -241,6 +238,10 @@ def _wait_for_gateway_ready(status_queue: queue.Queue[object], log_path: Path) -
 def provisioned_gateway(
     gateway_serial_port, gateway_creds_dir, gateway_creds, gateway_log_path
 ):
+    if not gateway_serial_port:
+        logging.info("No gateway port configured; skipping gateway provisioning")
+        yield
+        return
```
mniestroj, line 242: How about initially depending only on `request` and `gateway_serial_port`, return when the port is absent, and request the gateway-only fixtures dynamically after that check..

examples/zephyr/http_client/sample.yaml
```diff
@@ -38,7 +27,12 @@ tests:
       pytest_dut_scope: module
       pytest_root:
         - pytest/test_sample.py
+    timeout: 600
```
mniestroj: The HIL scenario gives the whole pytest subprocess 600 seconds, but its sequential waits can total approximately 1,140 seconds before flash and API overhead: 60 seconds for DUT readiness, 240 seconds for settings, 120 seconds for stream polling, and 720 seconds across the OTA waits.

A slow but valid device can remain within each individual timeout yet be terminated by Twister's outer timeout. Please either raise the scenario timeout above the combined worst-case budget or reduce/share the inner timeout budgets.

examples/zephyr/http_client/pytest/README.md
```diff
@@ -0,0 +1,85 @@
...
+fw_update you must build the update binary your self and supply the
+binary and version pytest args when calling twister.
+
+First, you must update the version number in the VERSION file:
+```
+# Update to your preferred version, like 2.0.0
+nano examples/zephyr/http_client/VERSION
+```
+
+The sample.yml sets the package name, you must use the same value. For
```
mniestroj:
```suggestion
The sample.yaml sets the package name, you must use the same value. For
```

examples/zephyr/http_client/pytest/README.md
```diff
@@ -0,0 +1,85 @@
...
+This directory contains local pytest-based HIL tests for the
+`examples/zephyr/http_client` sample.
+
+## Prerequisites
+
+1. Activate ESP-IDF environment:
+
+```bash
+. /path/to/esp-idf/export.sh
+```
```
mniestroj: We do not need ESP-IDF environment for Zephyr sample, right?

scripts/pytest-pouch/pytest_pouch/plugin.py
szczys: Great catch, thank you! When fixing this I also took this time to better handle the ota mode options so that they are defined in one place and used programmatically in the file.

scripts/pytest-pouch/pytest_pouch/gateway_provision.py
szczys: Ah, I didn't know about the dynamic call trick. This is nice to avoid generating the certs if we're not going to use them. I've updated as suggested, thank you!

examples/zephyr/http_client/sample.yaml
szczys: This is a good callout. I dug  through the cumulative tests and reduced them where possible. I added 120s of overhead to the cumulative (for firmware flashing, etc.) and made a new commit so we have a record of why these changes are being made.

scripts/pytest-pouch/pytest_pouch/gateway_provision.py
```diff
@@ -238,11 +235,17 @@ def _wait_for_gateway_ready(status_queue: queue.Queue[object], log_path: Path) -
 
 
 @pytest.fixture(scope="module", autouse=True)
-def provisioned_gateway(
-    gateway_serial_port, gateway_creds_dir, gateway_creds, gateway_log_path
-):
+def provisioned_gateway(request, gateway_serial_port):
+    if not gateway_serial_port:
+        logging.info("No gateway port configured; skipping gateway provisioning")
+        return
```
mniestroj-ai, line 242: The no-port branch returns before this generator fixture reaches a `yield`. Pytest therefore reports `ValueError: provisioned_gateway did not yield a value`, so the intended optional path fails during fixture setup whenever this plugin is loaded without `--gateway-port`.

Please retain a bare `yield` before `return`. The gateway-only fixtures can remain dynamically requested after this check.

examples/zephyr/http_client/pytest/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+```bash
+uv pip install -r requirements-ci-zephyr.txt --require-hashes
+```
+
+2. Install cloud tooling used by the test:
+
+```bash
+uv pip install "golioth@git+https://github.com/golioth/python-golioth-tools@v0.8.1"
+```
```
mniestroj-ai, line 18: This command runs `test_sensor_stream_uplink`, but the prerequisites do not mention the pipeline required to route `.s/sensor` JSON into LightDB Stream. Without that route, `device.stream.get()` finds no payload containing `temp` and the test times out even when the device is operating correctly.

Please add the equivalent pipeline prerequisite and configuration provided by the ESP-IDF HTTP client guide.

examples/esp_idf/http_client/pytest/README.md
```diff
@@ -55,49 +40,63 @@ export GOLIOTH_API_URL="https://api.golioth.io"
...
 export WIFI_SSID="your_wifi_ssid"
 export WIFI_PSK="your_wifi_psk"
+```
+
+## Build Update Firmware
 
-# Manual cert mode: provide DER file paths
-export DEVICE_CRT_DER_PATH="/path/to/device.crt.der"
-export DEVICE_KEY_DER_PATH="/path/to/device.key.der"
+Note: if you prefer to skip the fw_update test you can ignore this
```
mniestroj-ai, line 47: The non-OTA path says to ignore this entire section, but the separate normal firmware build section was removed and step 2 here is now the only instruction that builds the initial image. Following the documented non-OTA path therefore leaves no build directory for pytest.

Please retain an initial firmware build outside the OTA-only instructions, or clarify that non-OTA users skip only the update build and must still perform the initial build.

scripts/pytest-pouch/pytest_pouch/zephyr_harness.py
mniestroj: How about using such helper function, that will work with alternative `twister-out` directory names:
```python
def _twister_out_dir(device_object) -> Path:
    """Find the Twister output directory containing this test build."""
    build_dir = Path(device_object.device_config.build_dir)
    for candidate in (build_dir, *build_dir.parents):
        if (candidate / "testplan.json").is_file():
            return candidate
    pytest.fail(f"Could not locate Twister output directory above {build_dir}")
```

szczys: Ah, that's much better, thanks for the suggestion!

examples/esp_idf/http_client/pytest/README.md
szczys: The test command also builds the non-update version of the firmware. So there's no need to run an explicit build step. But to address your feedback I've added an additional subsection that shows how to run the test without the firmware update.

examples/zephyr/http_client/pytest/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+3. Enable the required Golioth pipeline route for stream validation:
+
+   The stream test (`test_sensor_uplink_contains_temp`) expects JSON data
```
mniestroj: Should it be `test_sensor_stream_uplink`?

examples/zephyr/http_client/pytest/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+### Run without Firmware Update Test
+```bash
+west twister -vv -W -T examples/zephyr/http_client                       \
+        --device-testing --device-serial /dev/ttyACM0                    \
+        --west-flash="--dev-id=001063461944"                             \
+        -p frdm_rw612                                                    \
+        --pytest-args="--generate-certs"
+```
```
mniestroj, line 99: Should we add `--pytest-args="--ota-mode=disabled"`?

mniestroj (approved):

PR #326, opened 2026-08-19
ci: capture and upload BLE gateway logs during HIL tests

mniestroj:
This is an alternative implementation to #323. It addresses the same flaky BLE HIL failures by:

1. Capturing raw gateway serial output for the complete pytest module and uploading it as `gateway-ble_gatt.log`
2. Waiting for `Scanning successfully started` before yielding the gateway fixture, which occurs only after initial cloud setup succeeds
3. Cleaning stale gateway logs before each run so failures cannot upload output from an earlier self-hosted runner job

Unlike #323, readiness is detected directly from incoming bytes instead of polling and decoding a growing log file. Reader failures are propagated to the fixture, split markers are handled across serial reads, timeout diagnostics are bounded to the final 200 bytes, and a context manager owns serial, file, and thread cleanup.

The workflow passes an explicit matrix-specific log path, preserves argument boundaries with a Bash array, and declares `pyserial` directly. Runtime behavior is validated by the existing ESP-IDF BLE HIL job.

### DIFF
--- a/.github/workflows/test-esp-idf-hil.yml
+++ b/.github/workflows/test-esp-idf-hil.yml
@@ -173,7 +173,10 @@ jobs:
       - name: Clean stale run assets
         run: |
           rm -rf "$SAMPLE_DIR/build"
-          rm -f "$UNITY_REPORT_NAME" "pytest-${{ matrix.name }}-hil.log"
+          rm -f \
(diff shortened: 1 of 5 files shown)


szczys (approved):
Looks great, thanks for the improvements!

PR #333, opened 2026-08-31
Change CI to run on `pull_request_target`

trond-snekvik:
With the "Require approval for all external contributors" flag set for the repo, and the token permissions being set to `content: read` and `actions: write`, I believe this should be perfectly safe for us, and allows fork PRs to run CI - automatically for us and upon approval for external contributors.

### DIFF
--- a/.github/workflows/ci-trigger.yml
+++ b/.github/workflows/ci-trigger.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request:
+  pull_request_target:
   workflow_dispatch:
 
 permissions:


.github/workflows/ci-trigger.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request:
+  pull_request_target:
```
szczys, line 7: I don't think it is safe to use this trigger for all of the workflows that this kicks off. The build step for untrusted branches should run in an unprivileged job which I think is what `pull_request` accomplishes. Let's discuss this synchronously to hash through some of the downsides.

szczys (changes requested):

mniestroj:
`pull_request_target` supplies the trusted base-repository revision instead of the PR merge revision. None of the eight reusable workflows specifies a checkout `ref`, so lint, builds, unit tests, and HIL all test `main`, not the proposed changes.

GitHub explicitly documents that `pull_request_target` workflows [always run regardless of fork-workflow approval settings](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories). External contributors could trigger the full suite without approval, including the self-hosted HIL jobs.

PR #345, opened 2026-09-08
Separate uplink sessions from pouches

trond-snekvik:
Today, one call to pouch_uplink_start() does two things at once: it starts a new crypto session (crypto_session_start()) and starts the first (and only) pouch (crypto_pouch_start()). pouch_uplink_finish() symmetrically closes both the pouch and the crypto session together. There is exactly one pouch per session, and the transport has no way to send a second pouch without tearing down and re-deriving the session key.

We want a transport to control the session boundary based on its own connection lifetime - one session per TLS connection (HTTP), one session per BLE connection to a gateway, or one long-running session for serial (until power-down or pouch-index exhaustion) - while sending multiple uplink pouches through that same session. This requires separating the session lifecycle from the pouch lifecycle, and propagating that separation through the transports.

The core part of this change is the new pouch_open/pouch_close API in include/pouch/transport/uplink.h, which has to be called within a valid uplink session (the old uplink_start/uplink_finish API).


port/zephyr/transport/gatt/peripheral.c
```diff
@@ -215,6 +216,14 @@ static void ccc_changed(const struct bt_gatt_attr *ccc_attr, uint16_t value)
     c->attr = NULL;
 }
 
+static void disconnected(struct bt_conn *conn, uint8_t reason)
+{
+    pouch_uplink_finish();
```
mniestroj: `BT_CONN_CB_DEFINE` fires for *every* connection. An application with any other BLE connection has its pouch session torn down by an unrelated disconnect.

mniestroj (commented):
Pouch number 65537 in a session gets `id == 1` again, and with `block_index` also restarting at 0 it reproduces pouch number 1's nonce byte-for-byte under the same key.

Should we enforce overflow with something like:
```
int saead_uplink_pouch_start(void)
{
    if (uplink.pouch.id == UINT16_MAX) {
        return -EOVERFLOW;
    }
    return session_pouch_start(&uplink, uplink.pouch.id + 1);
}
```
?

port/zephyr/transport/gatt/peripheral.c
trond-snekvik: Good point! I'll move this below the characteristics listing and compare it to `uplink.conn`, I think, but I'll do another pass to make sure that works.

trond-snekvik:
@mniestroj I believe this is the issue fixed by https://github.com/golioth/pouch/pull/315. In hindsight, #335 should have been rebased on that PR to show both changes working together

mniestroj:
> @mniestroj I believe this is the issue fixed by #315. In hindsight, #335 should have been rebased on that PR to show both changes working together

#315 guards against a different field of the same struct:
```
nonce[0..1] = be16(pouch.id)       <- #315 does not touch this
nonce[2..3] = be16(block_index)    <- #315's finding, guarded.
```

trond-snekvik:
Damn, you're right, I got my indexes mixed up! Pushed a separate commit to address this.

port/zephyr/transport/gatt/peripheral.c
```diff
@@ -273,3 +274,16 @@ BT_GATT_SERVICE_DEFINE(pouch,
...
 #endif
 );
+
+
+static void disconnected(struct bt_conn *conn, uint8_t reason)
+{
+    if (conn == uplink.conn)
+    {
+        pouch_uplink_finish();
+    }
```
mniestroj: Is that ever satisfied (is `(conn == uplink.conn)` true) if we have `c->conn = NULL;` above (line 215)?

mniestroj (approved):
The only outstanding comment is the one about `(conn == uplink.conn)`. Otherwise LGTM!

port/zephyr/transport/gatt/peripheral.c
trond-snekvik: Dang, true. I actually think we're a bit too sloppy with the connections here in general, and need more than just a quick fix. As the connection pointer is stored per characteristic, we could end up with different gateways interacting with different characteristics in parallel, which breaks the singleton approach further up the stack anyway. 

I'll revert this change, making us going back to misbehaving if there are multiple concurrent connections (which is already the case in a way), and open a separate PR that prevents multiple gateways from interacting with the pouch device simultaneously.

PR #349, opened 2026-09-09
examples: linux: rpmsg gateway (Go broker for Pouch Serial)

beriberikix:
The Linux half of the rpmsg transport: a Go daemon that brokers Pouch sessions
between the MCU on `/dev/rpmsgN` and Golioth over mTLS. This is what made the
end-to-end hardware validation of #304 possible.

**We expect this to end up in a production gateway rather than here.** It is
proposed as an example so the rpmsg transport has a working host counterpart to
be reviewed and validated against — which is what it has already done, carrying
the end-to-end hardware runs behind #304, #350 and #351 — but a gateway that
ships is not an example, and this would be better moved than maintained in two
places.

`internal/serial` is the load-bearing part and the piece worth moving. It is the
only implementation of Pouch Serial outside the firmware tree, and any host
gateway needs one regardless once SPI (#256) lands. It transfers roughly intact:
a pure state machine that never touches the link, so it carries no assumption
about rpmsg. `cmd/` and `internal/link` are the example-shaped parts and would be
dropped or replaced.

Merging here is therefore worth treating as provisional. If the move happens
soon it may be better not to land this at all, and to take `internal/serial`
straight across. See the discussion on #274.

Stack:
2. **this PR** `transport/rpmsg-broker` — the Go gateway
3. `transport/rpmsg-ota` — MCU-mediated firmware relay
4. `transport/rpmsg-signed-url` — signed-URL firmware delivery

### DIFF
--- a/examples/linux/rpmsg_gateway/README.md
+++ b/examples/linux/rpmsg_gateway/README.md
@@ -1,0 +1,106 @@
+# Pouch rpmsg Gateway
+
+The Linux half of Pouch on an MPU+MCU SoC. The MCU runs Zephyr and the Pouch
+device stack (see `examples/zephyr/rpmsg_device`); this daemon runs on the
+application processor, brokers the MCU's sync sessions to a Pouch-compatible
+server, and speaks the Pouch Serial protocol over an rpmsg endpoint.
+
+```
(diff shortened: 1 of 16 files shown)


beriberikix:

PR #350, opened 2026-09-09
transport: serial: MCU-mediated firmware relay

beriberikix:
A core that remoteproc loads from the host filesystem at every boot has no flash
of its own to write an update into, but it is still the authenticated Pouch
endpoint. It downloads its own image through Pouch OTA and streams the plaintext
to the host over two serial channels; the host verifies the SHA-256 from the
manifest, installs alongside the previous image and restarts the core.

Gated on `CONFIG_POUCH_SERIAL_FW_RELAY`, default n — the relay buffer is 8 KB of
static RAM that only remoteproc-loaded cores can use. The channel ids stay
unconditional so every build agrees on the wire format.

Validated end to end on an FRDM-IMX95 against production Golioth: the relayed
image arrives byte-identical to the cloud artifact, and the rejection path
retries three times and then lets the cloud re-offer, all within one boot.

Stack:
2. #349 `transport/rpmsg-broker` — the Go gateway
3. **this PR** `transport/rpmsg-ota` — MCU-mediated firmware relay
4. `transport/rpmsg-signed-url` — signed-URL firmware delivery

### DIFF
--- a/examples/zephyr/rpmsg_device/CMakeLists.txt
+++ b/examples/zephyr/rpmsg_device/CMakeLists.txt
@@ -11,3 +11,7 @@ target_sources(app PRIVATE
   src/main.c
   src/credentials.c
 )
+
+if(CONFIG_POUCH_SERIAL_FW_RELAY)
+  target_sources(app PRIVATE src/fw_relay.c)
+endif()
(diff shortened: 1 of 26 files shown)


beriberikix:

PR #351, opened 2026-09-09
transport: signed-URL firmware delivery for the rpmsg stack

beriberikix:
A flashless core has the host apply its updates. The relay in #350 does that by
pulling the image through Pouch and pushing the plaintext back out — every byte
crosses the on-die link twice, and an 8 KB buffer sits in MCU RAM to cover the
gap between the two directions.

This moves the authority instead of the bytes. The MCU holds the device key, so
it signs a URL for its own artifact with [`signy`][signy] and hands the host
that; the host fetches the image itself, verifies size and SHA-256 against the
manifest, installs it and restarts the core. Nothing is buffered on the MCU and
no image byte crosses the link.

Background: [Signed URLs for Embedded Devices][blog1] and [Signed URLs on
ESP32][blog2].

[signy]: https://github.com/golioth/signy
[blog1]: https://blog.golioth.io/signed-urls-for-embedded-devices/
[blog2]: https://blog.golioth.io/signed-urls-on-esp32/

## What is here

- **Verdict split out of the relay** so a second delivery mechanism can report
  results without carrying the relay's buffer.
- **Two channels**: `TIME` (7, host → device) and `FW_URL` (8, device → host),
  plus an INFO capability bit so a host that does not implement them never
  pushes to a device that cannot answer.
- **Device endpoints** behind `CONFIG_POUCH_SERIAL_FW_SIGNED_URL`, default n and
  independent of the relay. Signing is done through a callback the application
  installs, so `src/transport` never depends on the signing library.
- **Example** builds both mechanisms and chooses per attempt, falling back to
  the relay when a signed URL is refused.
- **Gateway** fetches the artifact in the background and reports the verdict on
  the next session; `-signed-url`, `-download-dir`, `-download-timeout`, `-ca`.
- **CI**: a Go job (nothing built the gateway before) and a second imx95 entry
  with the handoff off, so the relay-only configuration keeps compiling.

## Three decisions worth reviewing

**Signing is lazy.** A signed URL is valid for ~90 s and the host may not
collect for several sessions, so the signature is made when the transfer opens,
not when the artifact is announced.

**Endpoints never fail a transfer.** No signer, no clock yet, an over-long
signature: all produce an empty transfer with the handoff still pending. An
error would fail the whole session, which is also carrying the uplink and
downlink.

**Time gets its own channel.** The MCU has no wall clock. The verdict channel is
silent until there is a verdict, INFO runs the wrong way, and the certificate
channels stop once provisioned — so none of them can carry it. Being told the
time doubles as the device's evidence that the host speaks this protocol.

## Known wart

signy needs a wall clock, which pulls in Zephyr's POSIX timers, which puts the
POSIX headers on the include path, at which point mbedTLS finds an
`<arpa/inet.h>` that declares its functions with `sa_family_t` — a type only
defined when the networking stack is built. A Pouch device over rpmsg has no IP
stack, so `x509_crt.c` fails to compile. The example aliases the type so the
declaration parses; nothing calls `inet_pton()`. Reported upstream as golioth/signy#18: the library reads the clock itself, and
on Zephyr 4.4 every route to `clock_gettime()` reaches `POSIX_SYSTEM_INTERFACES`,
so letting the application supply the time is the fix that actually removes the
constraint.

## Testing

- 219 ztest cases pass on native_sim, 13 of them a new `pouch.serial.fw_url`
  suite covering the record layout, lazy signing, the unsignable-stays-pending
  path, and re-announcing from the verdict callback.
- Go tests for record parsing, the downloader against `httptest` (verified,
  corrupted, short, over-long, refused, length mismatch, cancelled), and three
  signed-URL exchanges through the broker state machine.
- imx95_evk/mimx9596/m7 builds in three configurations: both mechanisms, relay
  only, and native_sim.
- **Validated on an FRDM-IMX95 against production Golioth.** Device reports
  1.0.1, cloud offers 1.0.3, device signs the URL, gateway downloads and
  verifies 126580 bytes in about a second, verdict `ok`, core restarted, and the
  cloud then sees the device reporting 1.0.3. The relay fallback was exercised
  too, in the run where the signature was refused: device took the verdict,
  switched mechanism, and the update still completed.

## One finding worth knowing

Golioth enforces the not-before stamp with **no clock-skew tolerance at all**.
Signing URLs locally at a range of offsets against production:

| `nb` offset | Result |
|---|---|
| +0s | 200 |
| +5s | 401 |
| +20s | 401 |
| +60s | 401 |

The bench board's clock was twenty seconds fast, so every signed URL came back
401 — and because the device falls back to relaying, the update still succeeded
and the failure was invisible in everything except the gateway log.

So the gateway reports **the server's** clock to the device rather than its own.
The offset comes from the `Date` header of responses it already makes, including
the error responses seen before it can authenticate, so it costs no extra
request and is re-learned every session rather than fixed at startup. Gating the
feature on the two clocks agreeing was the alternative and is worse: it makes
the degradation loud but still gives up signed URLs over something correctable.

A few seconds of backdating remains on top, covering what correction cannot —
the header's one-second resolution and the transit time before the device signs.
A host that disagrees with the server is warned about anyway, since a wrong
local clock breaks certificate validity windows and log timestamps regardless.

Verified on hardware with the board's clock set **45 seconds fast**, far beyond
anything a fixed margin could absorb:

```
WARN  this host's clock disagrees with the server; signed URLs are corrected
      for it, but the host clock is worth fixing  behind_server_by=-43.857s
INFO  artifact downloaded and verified  package=rpmsg_device version=1.0.3
      bytes=126580
INFO  core restarted on the new image  firmware=rpmsg_device-1.0.3.elf
```

Stack:
2. #349 `transport/rpmsg-broker` — the Go gateway
3. #350 `transport/rpmsg-ota` — MCU-mediated firmware relay
4. **this PR** `transport/rpmsg-signed-url` — signed-URL firmware delivery

Refs #274.

### DIFF
--- a/.github/workflows/ci-trigger.yml
+++ b/.github/workflows/ci-trigger.yml
@@ -31,6 +31,10 @@ jobs:
       actions: write
     uses: ./.github/workflows/test-build-gateway.yml
     secrets: inherit
+  trigger-go:
+    name: Run Go Gateway Tests
+    uses: ./.github/workflows/test-go.yml
+    secrets: inherit
   trigger-twister:
     name: Run Twister
     permissions:
--- a/.github/workflows/test-build.yml
+++ b/.github/workflows/test-build.yml
@@ -30,6 +30,14 @@ jobs:
             example: rpmsg_device
             board: imx95_evk/mimx9596/m7
             no_sysbuild: true
+          # The same sample with the signed-URL handoff off, so the relay-only
+          # configuration keeps building: the two are independent, and only
+          # this entry compiles the fallback path on its own.
+          - manifest: west-zephyr.yml
+            example: rpmsg_device
+            board: imx95_evk/mimx9596/m7
+            no_sysbuild: true
+            extra_args: -DCONFIG_EXAMPLE_FW_SIGNED_URL=n
       fail-fast: false
     runs-on: ubuntu-24.04
     steps:
@@ -63,7 +71,9 @@ jobs:
             SYSBUILD_FLAG="--sysbuild"
           fi
 
-          west build ${SYSBUILD_FLAG} -p -b ${{ matrix.board }} pouch/examples/zephyr/${{ matrix.example }}
+          west build ${SYSBUILD_FLAG} -p -b ${{ matrix.board }} \
+            pouch/examples/zephyr/${{ matrix.example }} \
+            -- ${{ matrix.extra_args }}
 
   build-esp-idf:
     strategy:
--- a/.github/workflows/test-go.yml
+++ b/.github/workflows/test-go.yml
@@ -1,0 +1,48 @@
+name: "Test: Go Gateway"
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  go:
+    name: Vet, build and test the Go gateway
+    runs-on: ubuntu-24.04
+    defaults:
+      run:
+        working-directory: examples/linux/rpmsg_gateway
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Set up Go
+        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
+        with:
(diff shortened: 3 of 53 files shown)


beriberikix:

PR #353, opened 2026-09-09
Gateway: Ensure downlink is always closed

trond-snekvik:
While working on multi pouch sessions, I found that the gateway would leave the downlink open indefinitely if it never received any segments. This can happen both when there's no uplink data (which short circuits `send_uplink_via_cloud` in gateway/uplink.c) and when the downlink data is empty.

To deal with this without introducing any new API functions, we'll make two changes to the gateway behavior:

1. If the uplink is never started for any reason, we'll call the `pouch_gateway_downlink_end_cb` callback the same way we do after `pouch_gateway_cloud_forward_pouch` ends.
2. When ending the downlink, we'll now pass a synthetic empty last downlink block to `pouch_gateway_downlink_block_cb` to trigger the downlink_get_data machinery the same way the data receive callback would. Doing this instead of adding another "downlink_complete" function ensures that the downlink close processing is always done in the downlink endpoint context instead of running directly in the uplink close context.

I believe this issue is present even without the modified device behavior, we just never send empty uplinks in our applications.

### DIFF
--- a/src/gateway/downlink.c
+++ b/src/gateway/downlink.c
@@ -130,6 +130,14 @@ void pouch_gateway_downlink_end_cb(int status, void *arg)
                 downlink->data_available_cb(downlink->cb_arg);
             }
         }
+        return;
+    }
+
+    if (downlink->last_block == NULL)
+    {
+        // We never received the last block (or likely any data at all).
+        // Pass an empty dummy last block down to let the normal end-of-downlink mechanism run:
+        pouch_gateway_downlink_block_cb(NULL, 0, true, downlink);
     }
 }
 
--- a/src/gateway/uplink.c
+++ b/src/gateway/uplink.c
@@ -41,8 +41,16 @@ struct pouch_gateway_uplink
     void *end_cb_arg;
 };
 
-static void cleanup_uplink(struct pouch_gateway_uplink *uplink)
+static void end_uplink(struct pouch_gateway_uplink *uplink, int err)
 {
+    if (uplink->downlink)
+    {
+        pouch_gateway_downlink_end_cb(err, uplink->downlink);
+    }
+
+    uplink->end_cb(uplink->end_cb_arg,
+                   err ? POUCH_GATEWAY_UPLINK_ERROR_CLOUD : POUCH_GATEWAY_UPLINK_SUCCESS);
+
     struct pouch_buf *block;
 
     if (uplink->drain_block != NULL)
@@ -148,16 +156,14 @@ static void send_uplink_via_cloud(struct pouch_gateway_uplink *uplink)
     if (uplink->total_len == 0)
     {
         POUCH_LOG_DBG("No uplink data to send");
-        uplink->end_cb(uplink->end_cb_arg, POUCH_GATEWAY_UPLINK_SUCCESS);
-        cleanup_uplink(uplink);
+        end_uplink(uplink, 0);
         return;
     }
 
     if (!IS_ENABLED(CONFIG_POUCH_GATEWAY_CLOUD))
     {
         POUCH_LOG_INF("Cloud disabled, discarding %zu bytes of peripheral data", uplink->total_len);
-        uplink->end_cb(uplink->end_cb_arg, POUCH_GATEWAY_UPLINK_SUCCESS);
-        cleanup_uplink(uplink);
+        end_uplink(uplink, 0);
         return;
     }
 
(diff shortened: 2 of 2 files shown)


src/gateway/downlink.c
```diff
@@ -130,6 +130,14 @@ void pouch_gateway_downlink_end_cb(int status, void *arg)
                 downlink->data_available_cb(downlink->cb_arg);
             }
         }
+        return;
+    }
+
+    if (downlink->last_block == NULL)
```
mniestroj, line 136: The downlink has two users running in different execution contexts:

1. The gateway work queue owns `uplink->downlink` while forwarding the uplink to the cloud.
2. The BLE sender consumes and eventually closes the same downlink.

There is currently no ownership mechanism that keeps the downlink alive until the gateway work queue has finished using it.

The problematic sequence is:

**Gateway work queue:**

```text
close_work_handler()
  pouch_gateway_uplink_close()
    send_uplink_via_cloud()
      pouch_gateway_cloud_forward_pouch()
        pouch_coap_blockwise_post_streaming()
          pouch_coap_process_block2_response()
            cloud_downlink_cb(..., is_last=true)
              pouch_gateway_downlink_block_cb()
```

`pouch_gateway_downlink_block_cb()` stores the final block and wakes the BLE sender:

```c
downlink->last_block = block;
pouch_msgq_put(&downlink->block_queue, ...);
downlink->data_available_cb(...);
```

The BLE sender can now consume that block:

```text
pouch_bearer_ready()
  pouch_sender_ready()
    push_fragments()
      broker_endpoint_downlink.send()
        pouch_gateway_downlink_get_data()
```

Because this is the final block, `pouch_gateway_downlink_get_data()` marks the downlink complete:

```c
pouch_atomic_set_bit(downlink->flags, DOWNLINK_FLAG_COMPLETE);
*is_last = true;
```

The final BLE fragment is then transmitted. Before the cloud forwarding function has returned to `send_uplink_via_cloud()`, the BLE event thread can process the final ACK or a disconnect:

```text
BLE ACK/disconnect
  pouch_sender_recv() / pouch_sender_close()
    end(sender, ...)
      broker_endpoint_downlink.end()
        pouch_gateway_downlink_abort(node->downlink_ctx)
```

Since `DOWNLINK_FLAG_COMPLETE` is already set, `pouch_gateway_downlink_abort()` immediately destroys the context:

```c
if (pouch_gateway_downlink_is_complete(downlink))
{
    pouch_gateway_downlink_close(downlink);
}
```

`pouch_gateway_downlink_close()` eventually calls:

```c
free(downlink);
```

The gateway work queue still has that pointer in `uplink->downlink`. When cloud forwarding subsequently returns, it executes:

```text
pouch_gateway_cloud_forward_pouch() returns
  send_uplink_via_cloud()
    end_uplink()
      pouch_gateway_downlink_end_cb(0, uplink->downlink)
```

The pointer passed to `pouch_gateway_downlink_end_cb()` can now be dangling. This new line dereferences it:

```c
if (downlink->last_block == NULL)
```

In timeline form:


```text
Gateway work queue                         BLE event thread
------------------                         ----------------
Deliver final cloud block
Queue it in the downlink
Wake the BLE sender
Send the final BLE fragment
Return from the response callback
                                           Receive final ACK/disconnect
                                           Mark sender finished
                                           Abort completed downlink
                                           free(downlink)
Cloud forwarding returns
end_uplink()
downlink_end_cb(stale pointer)
read downlink->last_block  <-- use-after-free
```

The lifetime relationship was already fragile, but before this PR the successful `pouch_gateway_downlink_end_cb(0, ...)` path did not access `*downlink`.

trond-snekvik: This is valid on paper, and should be addressed, but in practice, I think it's just the abrupt disconnect that actually can happen, as the underlying transport is unlikely to even have time to schedule the TX for that message before the call returns. And I think that might have already been a problem here and other places where the work queue references the downlink pointer. 

We could add a mutex around the structure here, and arguably, it doesn't need to be allocated and freed for each connection, but I don't think this is a regression in this PR, so I would like to move that to an issue and address it separately. Thoughts?

mniestroj: Looking more at that, I think your proposal makes sense. We can always create new reproduction test and try to address that with future improvements.

mniestroj (approved):

src/gateway/downlink.c
trond-snekvik: Created an issue here: https://github.com/golioth/firmware-issue-tracker/issues/1077

PR #357, opened 2026-09-16
HIL tests: Increase auto negotiation timeout

trond-snekvik:
This is an experiment/shot in the dark to improve the reliability of the OTA HIL test, which is pretty flaky. This test appears to fail the autonegotiation to establish the physical layer for the ethernet connection, which times out after 4 seconds. By increasing this to 30 seconds, I hope to give whatever is non-responsive a few more tries before it gives up. When this fails, we're still spending 30 seconds waiting for the network interface to realize it has failed, so this should not change the time to failure for the test, as the interface doesn't appear to crash out early when the phy fails, but instead just sits there idling.


examples/zephyr/gateway/boards/frdm_rw612.conf
```diff
@@ -1,4 +1,7 @@
 CONFIG_NET_DHCPV4=y
 CONFIG_BUILD_OUTPUT_HEX=y
 
+# Give the PHY more time to complete Ethernet auto-negotiation on HIL runners.
+CONFIG_PHY_AUTONEG_TIMEOUT_MS=30000
```
trond-snekvik: what the heck

trond-snekvik:
The targeted HIL test has passed 4/4 runs with the new timeout in place. That was fairly unlikely without this fix, so I believe it's having the intended effect, but it's currently localized to the only test that was properly flaky. Would it be better to apply this some other place?

mniestroj (approved):

